PR #32, opened 2014-03-05
Brought rrig's implementation of kerberos delegation up to date with master

JoshuaWarburton:
Made the changes as suggested here a8a6ada1c143e5b339480b3d509b456a619ffae3 and merged with master

### DIFF
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -110,9 +110,9 @@ class HTTPKerberosAuth(AuthBase):
         try:
             result, self.context[host] = kerberos.authGSSClientInit(
                 "{0}@{1}".format(self.service, host), gssflags=gssflags)
-        except kerberos.GSSError:
+        except kerberos.GSSError as error:
             log.error("generate_request_header(): authGSSClientInit() failed:")
-            log.exception(e)
+            log.exception(error)
             return None
 
         if result < 1:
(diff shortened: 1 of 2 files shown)


JoshuaWarburton:
Is there anything more that needs to be done to merge this, or are we just waiting on approval?

requests_kerberos/kerberos_.py
```diff
@@ -96,11 +99,17 @@ def generate_request_header(self, response):
...
+        # Flags used by kerberos module.
+        gssflags = kerberos.GSS_C_MUTUAL_FLAG | kerberos.GSS_C_SEQUENCE_FLAG
+        if self.delegate:
+            gssflags |= kerberos.GSS_C_DELEG_FLAG
+
         try:
             result, self.context[host] = kerberos.authGSSClientInit(
-                "{0}@{1}".format(self.service, host))
+                "{0}@{1}".format(self.service, host), gssflags=gssflags)
         except kerberos.GSSError:
```
kevints: `except kerberos.GSSError as e:`?

JoshuaWarburton: Is that strictly necessary as we don't do anything with "e"?

kevints: You attempt to log it below

JoshuaWarburton: Ahh, yes. I was reading one except block too far down. Considering we log the exception here, it should probably be logged in the other blocks too.

JoshuaWarburton:
New commit to address comments from @kevints

mkomitee:
I've verified that the tests pass and that delegation works.

It also merges cleanly.

I'd really like to see this extended to use GSS_C_DELEG_POLICY_FLAG, but it's not available in kerberos (or pykerberos), so I don't think we should require it.

This looks good to me.

:+1:

sigmavirus24:
Thanks for the review @mkomitee

PR #33, opened 2014-06-17
Compatibility with Python 3

thomasgoirand:
Hi,

In the Debian package which I am about to upload, I'm embedding these 2 patches. It'd be nice to address the issue. I don't mind if you find another solution (especially for the test I disabled), but I would appreciate reactiveness. Best would be if you could get a new version out on PyPi ASAP with these fixes.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ def get_version():
     """
     reg = re.compile(r'__version__ = [\'"]([^\'"]*)[\'"]')
     with open('requests_kerberos/__init__.py') as fd:
-        matches = filter(lambda x: x, map(reg.match, fd))
+        matches = list(filter(lambda x: x, map(reg.match, fd)))
 
     if not matches:
         raise RuntimeError(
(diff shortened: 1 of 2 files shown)


mkomitee:
I'd be concerned with disabling that specific test because it's a security feature.

In that test, mutual authentication is required, an http error occurs, and mutual authentication is made to fail. 

When mutual authentication fails (when required) on otherwise successful responses, we raise an exception. When mutual authentication fails (when required) on an otherwise failed response, we return the response, but sanitize it to protect the user -- they wanted mutual authentication, it couldn't be provided, so they shouldn't get the response content.

sigmavirus24:
So the change to setup.py makes sense, but disabling that test doesn't. I'm going to cherry-pick the setup.py change and then close this.

sigmavirus24:
Oh I don't even have to do that: https://github.com/requests/requests-kerberos/blob/5e8ae5ce3a9f6256069b772bf579a4c4f05cc559/setup.py#L38

PR #42, opened 2014-11-17
Prevent infinite recursion when server responds with 401 to an authentication attempt.

rbcarson:
I found that requests-kerberos recurses until it blows the stack in the case where:

1. Server responds with 401 Unauthorized.
2. requests-kerberos retries the request with an Authorization header.
3. Server responds with 401 again, because the authenticated user has not been granted access.

This pull request changes the behaviour such that the response to `handle_401()` is returned with no further processing.

Tested with SharePoint Server 2013.

### DIFF
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -245,13 +245,14 @@ class HTTPKerberosAuth(AuthBase):
             response.request.body.seek(self.pos)
 
         if response.status_code == 401:
+            # 401 Unauthorized. Handle it, and if it still comes back as 401,
+            # that means authentication failed.
             _r = self.handle_401(response, **kwargs)
-            log.debug("handle_response(): returning {0}".format(_r))
-            return self.handle_response(_r, **kwargs)
         else:
             _r = self.handle_other(response)
-            log.debug("handle_response(): returning {0}".format(_r))
-            return _r
+            
+        log.debug("handle_response(): returning {0}".format(_r))
+        return _r
 
     def deregister(self, response):
         """Deregisters the response handler"""
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -238,21 +238,25 @@ class HTTPKerberosAuth(AuthBase):
 
     def handle_response(self, response, **kwargs):
         """Takes the given response and tries kerberos-auth, as needed."""
+        num_401s = kwargs.pop('num_401s', 0)
 
         if self.pos is not None:
             # Rewind the file position indicator of the body to where
             # it was to resend the request.
             response.request.body.seek(self.pos)
 
-        if response.status_code == 401:
+        if response.status_code == 401 and num_401s < 2:
             # 401 Unauthorized. Handle it, and if it still comes back as 401,
             # that means authentication failed.
             _r = self.handle_401(response, **kwargs)
+            log.debug("handle_response(): returning %s", _r)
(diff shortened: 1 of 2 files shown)


rbcarson:
@sigmavirus24 I got a chance to test the `fix-infinite-401` branch that you made (thanks!), at d23fc4427176aa73129a74deaf205ded114c0c04. Unfortunately, in the same original scenario, instead of blowing the stack, it now gets into a bad state:

```
(stack trace truncated)
  File "K:\code\install\python\requests\sessions.py", line 383, in request
    resp = self.send(prep, **send_kwargs)
  File "K:\code\install\python\requests\sessions.py", line 491, in send
    r = dispatch_hook('response', hooks, r, **kwargs)
  File "K:\code\install\python\requests\hooks.py", line 41, in dispatch_hook
    _hook_data = hook(hook_data, **kwargs)
  File "K:\code\install\python\requests_kerberos\kerberos_.py", line 255, in handle_response
    return self.handle_response(_r, num_401s=num_401s, **kwargs)
  File "K:\code\install\python\requests_kerberos\kerberos_.py", line 255, in handle_response
    return self.handle_response(_r, num_401s=num_401s, **kwargs)
  File "K:\code\install\python\requests_kerberos\kerberos_.py", line 257, in handle_response
    _r = self.handle_other(response)
  File "K:\code\install\python\requests_kerberos\kerberos_.py", line 181, in handle_other
    if not self.authenticate_server(response):
  File "K:\code\install\python\requests_kerberos\kerberos_.py", line 226, in authenticate_server
    _negotiate_value(response))
  File "K:\code\install\python\kerberos.py", line 218, in authGSSClientStep
    err, sec_buffer = context["csa"].authorize(data)
  File "K:\code\install\python\python27\win32\lib\sspi.py", line 139, in authorize
    sec_buffer_out)
error: (-2146893048, 'InitializeSecurityContext', 'The token supplied to the function is invalid')
```

This bad state only happens when `mutual_authentication` is `REQUIRED` or `OPTIONAL`. My theory for why this happens:

1. It appears that server authentication can only occur after client authentication.
1. Server authentication occurs in `handle_other()`.
1. Giving up and going to `handle_other()` after hitting 2 consecutive 401s results in an invalid Kerberos operation if server authentication is attempted. `_negotiate_value()` is probably returning `None`.

If you agree with this analysis, I'll proceed to work on a solution.

Confession: Instead of using the `kerberos` module, I'm using the API-equivalent `kerberos-sspi` module in order to avoid compiling binaries on Windows. I make `requests-kerberos` load it by renaming `kerberos_sspi.py`. This explains the final steps of the trace, though I don't believe this plays a role in the issue here.

sigmavirus24:
We probably want to just check the count before we try to handle a 401 or handle "other". Make sense?

sigmavirus24:
If you want to pull that branch into this one and continue working on it, @rbcarson that'd be awesome

rbcarson:
Okay, will do.

rbcarson:
This final check to prevent 401 responses from reaching `handle_other()` seems to eliminate the infinite recursion. Now, the 401 response is returned if the user is actually unauthorized, regardless of the `mutual_authentication` setting.

sigmavirus24:
Ah awesome. Thanks @rbcarson this makes total sense. I'm comfortable merging this. Ping me this weekend if I forget to do a release tomorrow.

I missed the fact that you updated this.

rbcarson:
Thanks @sigmavirus24. Does this mean #47 and #48 can be closed now? And please cut a release. :)

PR #46, opened 2015-02-24
Stop logging on otherwise acceptable responses

sigmavirus24:
cloudera/hue developers were seeing a large number of log errors for
200 status code responses. Instead of slowly growing a list of status codes
to not log an error message for, let's just not log the message if the response
code is "ok" (`200 <= status_code < 400`, vaguely speaking).


Lukasa:
LGTM. :cake:

PR #47, opened 2015-03-09
Fix infinite loop when run kdestory

BrickXu:
Refs #44.

### DIFF
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -158,6 +158,10 @@ class HTTPKerberosAuth(AuthBase):
         log.debug("handle_401(): Handling: 401")
         if _negotiate_value(response) is not None:
             _r = self.authenticate_user(response, **kwargs)
+            if _r is response:
+                log.debug("handle_401(): fail to authenticate, something wrong "
+                          "with the client.")
+                _r.raise_for_status()
             log.debug("handle_401(): returning {0}".format(_r))
             return _r
         else:


requests_kerberos/kerberos_.py
```diff
@@ -135,7 +135,8 @@ def authenticate_user(self, response, **kwargs):
         auth_header = self.generate_request_header(response)
         if auth_header is None:
             # GSS Failure, return existing response
-            return response
+            # NOTE: do not return origin response, it would cause infinite loop.
+            return None
```
sigmavirus24: You can return the original response. It will be unmodified so you can change the logic below.

requests_kerberos/kerberos_.py
```diff
@@ -158,6 +159,9 @@ def handle_401(self, response, **kwargs):
         log.debug("handle_401(): Handling: 401")
         if _negotiate_value(response) is not None:
             _r = self.authenticate_user(response, **kwargs)
+            if _r is None:
```
sigmavirus24: This can be

```py
if _r is response:
```

requests_kerberos/kerberos_.py
```diff
@@ -158,6 +159,9 @@ def handle_401(self, response, **kwargs):
...
+                log.debug("handle_401(): can not find any kerberos ticket.")
+                raise RuntimeError("Can not find any kerberos ticket.")
```
sigmavirus24: We shouldn't be raising a RuntimeError. We should be raising an exception from within the library.

requests_kerberos/kerberos_.py
tardyp: +1

requests_kerberos/kerberos_.py
tardyp: doc says that if auth fail we should raise HTTPError(401).
So I would be for:

```
if _r is response:
    _r.raise_for_status()
```

requests_kerberos/kerberos_.py
```diff
@@ -158,6 +159,9 @@ def handle_401(self, response, **kwargs):
         log.debug("handle_401(): Handling: 401")
         if _negotiate_value(response) is not None:
             _r = self.authenticate_user(response, **kwargs)
+            if _r is None:
+                log.debug("handle_401(): can not find any kerberos ticket.")
```
tardyp: There are 5 or 6 gss errors that are handled in previous code. not only the (arguably most common) ticket unavailable

tardyp:
There are actually 3 pull requests resolving the same issue :)

providing a 4th solution as a review comment to this one.

requests_kerberos/kerberos_.py
BrickXu: yes, you are right. I will fix it.

requests_kerberos/kerberos_.py
BrickXu: accept. +1

requests_kerberos/kerberos_.py
BrickXu: accept +1

BrickXu:
@tardyp @sigmavirus24 Thanks very much for reviewing, I've rebased the code.

rbcarson:
With these changes, I still see infinite recursion behaviour in the case where I use Kerberos to authenticate with SharePoint Server 2012, but my domain account is not authorized to access the site.

sigmavirus24:
This should be fixed in the next release by #42. If it isn't, please open a new issue.

PR #43, opened 2014-12-01
Fix infinite recursion

danc86:
Alternative attempt at fixing the infinite recursion problem (#42).

PR #48, opened 2015-03-17
Fix infinite recursion

erickt:
This updates @danc86's #43 patch to address @sigmavirus24's comment and resolve a merge conflict with the HEAD branch.


test_requests_kerberos.py
```diff
@@ -211,24 +211,6 @@ def test_authenticate_server(self):
             self.assertTrue(result)
             clientStep_complete.assert_called_with("CTX", "servertoken")
 
-    def test_handle_other(self):
```
sigmavirus24: Why does this need to be deleted? It seems handle_other was renamed, why not  rename things here appropriately?

sigmavirus24:
This should be resolved in the next release. If it isn't, please open a new issue.

PR #53, opened 2015-06-11
relax version in kerberos requirement

danc86:
Any version of the kerberos module should work fine, it has had no API changes in living memory as far as I can see.


sigmavirus24:
I think we want a lower-bound when we can confirm that it works for us.

PR #55, opened 2015-08-21
Uses kerberos_sspi on windows

### LINKED ISSUE #51: On Windows, 0.7.0 wheel still depends on kerberos module instead of kerberos_sspi
#34 changed setup.py to require kerberos_sspi instead of kerberos==1.1.1. However, the [wheel for 0.7.0](https://pypi.python.org/packages/2.7/r/requests-kerberos/requests_kerberos-0.7.0-py2.py3-none-any.whl#md5=384f5ee3e9922bc790feda79f0441538) still has a metadata.json containing:

```
"run_requires": [{"requires": ["requests (>=1.1.0)", "kerberos (==1.1.1)"]}]
```

So on Windows, `pip install requests-kerberos` still tries to install kerberos instead of kerberos_sspi.

AnthonyTruchet:
This is an attempt to fix #51.

It was done and validated on Windows aith Python 3.4 installed through anaconda.
I lack experience in compatibility issue between python 2 and python 3 : your guidance in how to reach an acceptable patch is more then welcome :-)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,7 @@ def get_version():
     """
     reg = re.compile(r'__version__ = [\'"]([^\'"]*)[\'"]')
     with open('requests_kerberos/__init__.py') as fd:
-        matches = filter(lambda x: x, map(reg.match, fd))
+        matches = list(filter(lambda x: x, map(reg.match, fd)))
 
     if not matches:
         raise RuntimeError(
--- a/test_requests_kerberos.py
+++ b/test_requests_kerberos.py
@@ -5,7 +5,15 @@
 
 from mock import Mock, patch
 import requests
-import kerberos
+
+
+try:
+    import kerberos
+    kerberos_module_name='kerberos'
+except ImportError:
+    import kerberos_sspi as kerberos # On Windows
+    kerberos_module_name='kerberos_sspi'
+
 import requests_kerberos
 import unittest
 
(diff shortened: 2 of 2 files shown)


rbcarson:
I'm not the maintainer of this library, but I can confirm that the unit tests now run on Windows under Python 2.7 thanks to these changes.

Lukasa:
This looks plausible to me. @rbcarson, do these tests run clean on your *nix platform?

rbcarson:
@Lukasa I'll have to get back to you on that, though I think you're in a better position to test that than me (isn't the Travis setup here using *nix?).

Lukasa:
It is, but I'd like a more direct live test if possible. =P

rbcarson:
@Lukasa The module installs successfully and the unit tests pass on Ubuntu 14.04 and Python 2.7.6. It fails to install on Python 3.4.0 because the `kerberos` module's setup.py isn't compatible with PY3.

`pykerberos`, a drop-in replacement for `kerberos` with the same module name, appears to work with PY3. The unit tests pass when I install pykerberos and skip requests-kerberos' setup.py. Perhaps it'd be safe to swap the dependency here?

Lukasa:
Ok, let's close this in favour of #56.

rbcarson:
@Lukasa We still need this PR's changes. #56 doesn't intersect with it.

Lukasa:
Heh, whoops: you're right. Clearly I'm doing too many things at once.

AnthonyTruchet:
I usually work in a corporate env where we use a review system and can rebase heavily to split commit. How should I proceed to split the kerberos_sspi part and the fix of the setup.py for Python 3 ? Or shall we keep it in the same commit ?

Lukasa:
For the moment I'm ok with leaving them as they are.

rbcarson:
I think this PR is good to merge. :+1:

Lukasa:
@sigmavirus24 You happy as well?

sigmavirus24:
I trust @rbcarson so yes.

PR #38, opened 2014-09-18
Switch requirement to pykerberos

Daviey:
Previously requirements.txt declared a requirement on
kerberos==1.1.1.  However, pykerberos is more active
upstream.  We also remove the version tracking as the
current minimum version on pypi is functional.

Therefore switching to that, which closes #37

Signed-off-by: Dave Walker (Daviey) <email@daviey.com>

PR #34, opened 2014-07-02
Add Windows native authentication support

gentoo90:
Add kerberos-sspi as an alternative backend.
[kerberos-sspi](https://github.com/may-day/kerberos-sspi) package is an API level equivalent with kerberos
package but uses windows sspi (through pywin32)

Issue #52, opened 2015-06-04
python3 and kerberos library


jamielennox:
The kerberos library appears essentially abandonded. Amongst other things there will be no python 3.

We should transition the library to use python-gssapi which is a higher level wrapper that uses cffi and is actively maitained. 

https://pypi.python.org/pypi/python-gssapi/
https://github.com/sigmaris/python-gssapi

Lukasa:
We agree that we should switch, but we've been unclear on what we should be switching to: see #38.

ASzc:
My Python 3 patch for PyKerberos was [merged](http://trac.calendarserver.org/ticket/798#comment:6) upstream a few months ago, and Apple released [kerberos 1.2.2](https://pypi.python.org/pypi/kerberos/1.2.2) on pypi, which includes it.

So, if you want to switch away from PyKerberos because of the slow updates, fair enough, but it does support Python 3 now.

dreyco676:
I'm working on a quick and dirty port of requests-kerberos to python3 which uses PyKerberos. Is there another library I should consider instead?

ASzc:
@dreyco676 , note that there's some ambiguity with the name PyKerberos. The original library produced by Apple is:

- called PyKerberos upstream (Issue Tracker, etc.), 
- called `kerberos` on PyPI,
- and often packaged as `python-kerberos` by Linux distros.

During the four years that PyKerberos went without an update, some unknown person uploaded a package called `pykerberos` to PyPI (with some improvements), leading a lot of people to mistake this for the original package.

Since then, the Apple version has released an update (version 1.2.2) that includes fixes, like my Python 3 patch. Since the PyPI `pykerberos` so far doesn't have a source repo or issue tracker, I would suggest staying with `kerberos`. They're only interchangeable because they're APIs happen to be the same, no other Kerberos libraries can function for requests-kerberos at the moment.

dreyco676:
Good lord, that was really nasty to get Apple's Kerberos compiled for x64 Windows. Will that always be the case or can we get the contributors from Apple to provide some binaries?

ASzc:
@dreyco676 It doesn't seem that Apple's maintainer (wsanchez) has much time to spare for PyKerberos, so contributions take a long time to get merged. I wouldn't expect any major improvements to happen at all without a patch being submitted.

[...]

PR #56, opened 2015-08-24
Fix dependency for kerberos-sspi on Windows; else use pykerberos. (Fixes #51, #38)

### LINKED ISSUE #51: On Windows, 0.7.0 wheel still depends on kerberos module instead of kerberos_sspi

rbcarson:
This change fixes the dependency issue in the whl. Now the metadata.json in the whl looks like:

```json
"run_requires": [
   {"environment": "sys_platform==\"win32\"", "requires": ["kerberos-sspi (>=0.2)"]}, 
   {"environment": "sys_platform!=\"win32\"", "requires": ["pykerberos (>=1.1.1,<2.0.0)"]}, 
   {"requires": ["requests (>=1.1.0)"]}
]
```

For PY3 support, I took the liberty of changing the *nix dependency to use `pykerberos` instead of `kerberos`, but I'm not able to test against a real Kerberos-authenticating server on *nix. I fixed the version based on the discussion in #38.

I set kerberos-sspi to >= 0.2 because 0.2 is the one that claims PY3 support, and based on the discussion in #34. It's also the only version that I've been using. In the highly unlikely (AFAIK) event that someone on Windows wants to install requests-kerberos and doesn't want to use requests-sspi because they have access to (py)kerberos, they can use `pip install --no-deps`...

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 requests>=1.1.0
 kerberos-sspi >= 0.2; sys.platform == 'win32'
-pykerberos >= 1.1.1, < 2.0.0; sys.platform != 'win32'
+pykerberos >= 1.1.8, < 2.0.0; sys.platform != 'win32'
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,3 @@
 requests>=1.1.0
+kerberos-sspi >= 0.2; sys.platform == 'win32'
+pykerberos >= 1.1.1, < 2.0.0; sys.platform != 'win32'
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ setup(
     ],
     extras_require={
         ':sys_platform=="win32"': ['kerberos-sspi>=0.2'],
-        ':sys_platform!="win32"': ['pykerberos>=1.1.1,<2.0.0'],
+        ':sys_platform!="win32"': ['pykerberos>=1.1.8,<2.0.0'],
     },
     test_suite='test_requests_kerberos',
     tests_require=['mock'],
--- a/setup.py
+++ b/setup.py
@@ -1,17 +1,8 @@
 #!/usr/bin/env python
 # coding: utf-8
 import os
-import sys
 import re
 from setuptools import setup
-
-with open('requirements.txt') as requirements:
-    requires = [line.strip() for line in requirements if line.strip()]
-
-if sys.platform == 'win32':
-    requires.append('kerberos-sspi')
-else:
-    requires.append('kerberos==1.1.1')
 
 path = os.path.dirname(__file__)
 desc_fd = os.path.join(path, 'README.rst')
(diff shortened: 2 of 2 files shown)


rbcarson:
@02strich It looks like we need a release of pykerberos that includes 02strich/pykerberos#5 to make Travis happy. :boom:

Lukasa:
Agreed, looks like we have to wait for this for a little bit. @rbcarson are you going to be able to keep an eye on the release of that project?

AnthonyTruchet:
May I ask why (after reading issue #38) you decided for pykerberos instead of a recent version of kerberos (the 1.2.2) ? Thanks for you reactivity by the way :-)

Lukasa:
Again, I'm going to let @rbcarson express a preference here =)

02strich:
@rbcarson just cut a release (1.1.8) of it

rbcarson:
Thanks, @02strich - looks like it works! 

@Lukasa, can we also have Travis test Python 3.4?

@AnthonyTruchet: I just noticed that the comments in #52 make a case for keeping kerberos at 1.2.2 due to added Python 3 support, but it still doesn't quite support it yet: its setup.py calls `file()`, which isn't in Python 3. And I don't know how to propose a change to fix that. @02strich's pykerberos, however, looks promising. I ruled out python-gssapi for this PR because it exposes a completely different API vs (py)kerberos. Do you have any preference here?

AnthonyTruchet:
I do not have any preference as I'm just discovering the Kerberos realm. I simply wanted to be sure the pieces of information in issue #52 had been taken into account. I tend to favor well established libraries when they are still maintained wrt to promising libraries with uncertain status. But I'm not sure the case is so clean here. 
By the way I have access to Python 2 and 3 on Windows (and Darwin with macports), let me know if I can help in any way.

Lukasa:
@rbcarson I've added 3.4 to the travis testing matrix on the master branch. For obvious reasons we won't release until it passes. ;)

mkomitee:
FWIW, I just pulled this in and tested it. It works in my kerberos environment (mit/linux), so pykerberos seems to be sufficient for the base-case.

rbcarson:
:clap: I think this is good to go, then. What do you think, @Lukasa?

sigmavirus24:
Closing and reopening to hopefully trigger travis to run against python 3.4

sigmavirus24:
So I'm going to merge this, but we learned over on [urllib3](/shazow/urllib3) that older versions of setuptools and pip will totally explode on having the extras defined this way for platform-specific dependencies. I'll fix these up such that they work well for Wheels and all versions of setuptools/pip.

PR #59, opened 2015-10-31
Translate extras to setup.cfg

sigmavirus24:
Fix the extras requirements for older setuptools/pip

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,7 @@
 [wheel]
 universal = 1
+
+[metadata]
+requires-dist =
+    kerberos-sspi>=0.2; :sys_platform=="win32"
+    pykerberos>=1.1.8,<2.0.0; :sys_platform!="win32"
--- a/setup.py
+++ b/setup.py
@@ -50,10 +50,6 @@ setup(
     install_requires=[
         'requests>=1.1.0',
     ],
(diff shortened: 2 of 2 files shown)


rbcarson:
What is there left to do to get this PR merged/closed?

behackett:
Is this just for pre-1.1 versions of pip? setuptools has supported extras_require since 0.5.

Lukasa:
I kicked the builders just to see if this is ok, but the builds were red before so I suspect this change isn't ready to go yet.

rbcarson:
There haven't been any complaints about the current setup, so is the goal of this PR worth the effort?

Lukasa:
That's for @sigmavirus24 to say, I think.

JensTimmerman:
the current setup doesn't work correctly when creating rpm's

```

====================================================================================================================================================================================================================================================================
 Package                                                                   Arch                                                    Version                                                       Repository                                                    Size
====================================================================================================================================================================================================================================================================
Updating:
 python-requests-kerberos                                                  noarch                                                  0.12.0-1                                                      homemade                                                      26 k
Removing for dependencies:
 python-kerberos                                                           x86_64                                                  1.1-15.el7                                                    @c7x_x86_64                                                   53 k


```

PR #69, opened 2016-04-12
Fix for mutual auth on winrm

### LINKED ISSUE #65: Add preemptive authentication
I'd like to use opportunistic authentication on the server end and simply check for WWW-Authenticate header presence. The main idea is to prevent the first HTTP request which will always return 401 and instead supply the Kerberos credentials immideately. At first glance it seems to be pretty easy to implement on the client side.

### LINKED ISSUE #68: Mutual auth REQUIRED fails w/ persistent connection
When communicating with a Windows wsman endpoint using requests-kerberos, we should be able to set mutual auth to REQUIRED. However, the server stops sending auth responses on persistent connections after a successful exchange, which causes requests-kerberos to fail for mutual auth on subsequent responses (because the WWW-Authenticate response header is missing). If authentication is  repeated on every request, the server will emit the proper headers in the response, but that behavior isn't currently implemented by requests-kerberos. 

I'm not familiar enough with the various RFCs to be sure, but I think this behavior falls into a gray area that's not well documented. 

I'm guessing one of two things should happen here:
1) Mutual auth REQUIRED forces reauthentication on each request (sub-optimal for performance, probably more so in delegated cases). A fix for this could also probably easily address #65.

2) Add stateful auth code to requests-kerberos that recognizes a successful auth exchange and won't try to validate mutual auth for implied-prior-auth responses on persistent connections. It'd still need to watch for 401s and reauthenticate accordingly (eg, proxies w/o Proxy-Support: Session-Based-Authentication), and it might need to be connection-aware to avoid a connection-hijacking attack via a compromised proxy (assuming requests surfaces what we'd need to know).

I can take a stab at implementing one of these behaviors, but I'm not sure which one is more correct. I'd lean toward 2) for performance, but implementing connection-aware behavior could be problematic. 1) should be more straightforward, but may have a performance penalty.

nitzmahone:
Fixes #68.

Existing code doesn't work with Windows Negotiate if mutual_auth is REQUIRED, since current code stops sending Authorization headers after success, but Windows keeps a persistent connection and stops sending WWW-Authenticate responses. This change allows Kerberos auth to be forced on each request by setting force_preemptive=True (defaults to False). 

Also "accidentally" fixes #65, since we needed a way to preemptively auth on every request anyway.

Updated existing tests to pass with changed internal args, but haven't added tests for new functionality yet. This is obviously "work in progress"- if the approach is sound, I'll add tests and clean up other things.

### DIFF
--- a/requests_kerberos/exceptions.py
+++ b/requests_kerberos/exceptions.py
@@ -10,3 +10,6 @@ from requests.exceptions import RequestException
 
 class MutualAuthenticationError(RequestException):
     """Mutual Authentication Error"""
+
+class KerberosExchangeError(RequestException):
+    """Kerberos Exchange Failed Error"""
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -11,10 +11,9 @@ from requests.compat import urlparse, StringIO
 from requests.structures import CaseInsensitiveDict
 from requests.cookies import cookiejar_from_dict
 
-from .exceptions import MutualAuthenticationError
+from .exceptions import MutualAuthenticationError, KerberosExchangeError
 
 log = logging.getLogger(__name__)
-
 
 # Different types of mutual authentication:
 #  with mutual_authentication set to REQUIRED, all responses will be
@@ -30,7 +29,6 @@ log = logging.getLogger(__name__)
 REQUIRED = 1
 OPTIONAL = 2
 DISABLED = 3
-
 
 class SanitizedResponse(Response):
     """The :class:`Response <Response>` object, which contains a server's
@@ -86,21 +84,22 @@ class HTTPKerberosAuth(AuthBase):
     object."""
     def __init__(
             self, mutual_authentication=REQUIRED,
-            service="HTTP", delegate=False):
+            service="HTTP", delegate=False, force_preemptive=False):
         self.context = {}
         self.mutual_authentication = mutual_authentication
         self.delegate = delegate
         self.pos = None
         self.service = service
+        self.force_preemptive = force_preemptive
 
-    def generate_request_header(self, response):
+    def generate_request_header(self, response, host, is_preemptive=False):
         """
         Generates the GSSAPI authentication token with kerberos.
 
-        If any GSSAPI step fails, return None.
+        If any GSSAPI step fails, raise KerberosExchangeError
+        with failure detail.
 
         """
-        host = urlparse(response.url).hostname
 
         # Flags used by kerberos module.
         gssflags = kerberos.GSS_C_MUTUAL_FLAG | kerberos.GSS_C_SEQUENCE_FLAG
@@ -108,48 +107,52 @@ class HTTPKerberosAuth(AuthBase):
             gssflags |= kerberos.GSS_C_DELEG_FLAG
 
         try:
+            kerb_stage = "authGSSClientInit()"
             result, self.context[host] = kerberos.authGSSClientInit(
                 "{0}@{1}".format(self.service, host), gssflags=gssflags)
-        except kerberos.GSSError as error:
-            log.error("generate_request_header(): authGSSClientInit() failed:")
-            log.exception(error)
-            return None
 
-        if result < 1:
-            log.error("generate_request_header(): authGSSClientInit() failed: "
-                      "{0}".format(result))
-            return None
+            if result < 1:
+                raise EnvironmentError(result, kerb_stage)
 
-        try:
+            # if we have a previous response from the server, use it to continue
+            # the auth process, otherwise use an empty value
+            negotiate_resp_value = '' if is_preemptive else _negotiate_value(response)
+
+            kerb_stage = "authGSSClientStep()"
             result = kerberos.authGSSClientStep(self.context[host],
-                                                _negotiate_value(response))
-        except kerberos.GSSError as error:
-            log.exception(
-                "generate_request_header(): authGSSClientStep() failed:")
-            log.exception(error)
-            return None
+                                                negotiate_resp_value)
 
-        if result < 0:
-            log.error(
-                "generate_request_header(): authGSSClientStep() failed: "
-                "{0}".format(result))
-            return None
+            if result < 0:
+                raise EnvironmentError(result, kerb_stage)
 
-        try:
+            kerb_stage = "authGSSClientResponse()"
             gss_response = kerberos.authGSSClientResponse(self.context[host])
+
+            return "Negotiate {0}".format(gss_response)
+
         except kerberos.GSSError as error:
             log.exception(
-                "generate_request_header(): authGSSClientResponse() failed:")
+                "generate_request_header(): {0} failed:".format(kerb_stage))
             log.exception(error)
-            return None
+            raise KerberosExchangeError("%s failed: %s" % (kerb_stage, str(error.args)))
 
-        return "Negotiate {0}".format(gss_response)
+        except EnvironmentError as error:
+            # ensure we raised this for translation to KerberosExchangeError
+            # by comparing errno to result, re-raise if not
+            if error.errno != result:
+                raise
+            message = "{0} failed, result: {1}".format(kerb_stage, result)
+            log.error("generate_request_header(): {0}".format(message))
+            raise KerberosExchangeError(message)
 
     def authenticate_user(self, response, **kwargs):
         """Handles user authentication with gssapi/kerberos"""
 
-        auth_header = self.generate_request_header(response)
-        if auth_header is None:
+        host = urlparse(response.url).hostname
+
+        try:
+            auth_header = self.generate_request_header(response, host)
+        except KerberosExchangeError:
             # GSS Failure, return existing response
             return response
 
@@ -285,6 +288,17 @@ class HTTPKerberosAuth(AuthBase):
         response.request.deregister_hook('response', self.handle_response)
 
     def __call__(self, request):
+        if self.force_preemptive:
+            # add Authorization header before we receive a 401
+            # by the 401 handler
+            host = urlparse(request.url).hostname
+
+            auth_header = self.generate_request_header(None, host, is_preemptive=True)
+
+            log.debug("HTTPKerberosAuth: Preemptive Authorization header: {0}".format(auth_header))
+
+            request.headers['Authorization'] = auth_header
+
         request.register_hook('response', self.handle_response)
         try:
             self.pos = request.body.tell()
--- a/README.rst
+++ b/README.rst
@@ -62,6 +62,19 @@ authentication, you can do that as well:
     >>> r = requests.get("http://example.org", auth=kerberos_auth)
     ...
 
+Preemptive Authentication
+-------------------------
+
+``HTTPKerberosAuth`` can be forced to preemptively initiate the Kerberos GSS exchange and present a Kerberos ticket on the initial request (and all subsequent). By default, authentication only occurs after a ``401 Unauthorized`` response containing a Kerberos or Negotiate challenge is received from the origin server. This can cause mutual authentication failures for hosts that use a persistent connection (eg, Windows/WinRM), as no Kerberos challenges are sent after the initial auth handshake. This behavior can be altered by setting  ``force_preemptive=True``:
+
+.. code-block:: pycon
+    
+    >>> import requests
+    >>> from requests_kerberos import HTTPKerberosAuth, REQUIRED
+    >>> kerberos_auth = HTTPKerberosAuth(mutual_authentication=REQUIRED, force_preemptive=True)
+    >>> r = requests.get("https://windows.example.org/wsman", auth=kerberos_auth)
+    ...
+
 Logging
 -------
 
--- a/test_requests_kerberos.py
+++ b/test_requests_kerberos.py
@@ -4,6 +4,7 @@
 """Tests for requests_kerberos."""
 
 from mock import Mock, patch
+from requests.compat import urlparse
 import requests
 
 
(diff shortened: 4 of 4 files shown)


requests_kerberos/kerberos_.py
```diff
@@ -285,6 +290,22 @@ def deregister(self, response):
...
+            # add Authorization header before we receive a 401
+            # by the 401 handler
+            host = urlparse(request.url).hostname
+
+            auth_header = self.generate_request_header(None, host, is_preemptive=True)
+
+            if auth_header is None:
+                # GSS setup failure, return error
+                # TODO: capture error detail and return in a typed exception?
+                raise Exception("gss setup broke")
```
nitzmahone: Any preference on how we fail here? We don't have a response yet...

Lukasa: Urgh, I'd say we should define a new exception for this case.

Lukasa:
This looks reasonable enough to me, modulo the outstanding work on the exception. Ping me again when you've dealt with that?

nitzmahone:
Right on, will do. Thanks for the look...

nitzmahone:
@Lukasa - ready for review.

requests_kerberos/kerberos_.py
```diff
@@ -11,10 +11,29 @@
...
+    def __enter__(self, **kwargs):
+        self.log_to_sniff.addHandler(self)
+        return self
+
+    def __exit__(self, type, value, traceback):
+        self.log_to_sniff.removeHandler(self)
+
+    def emit(self, record):
+        self.stored_logs.append(record.getMessage())
+
```
Lukasa: This is a really clever piece of code, but I'm disinclined to use it.

My particular worries are around having multiple co-operating threads (green or otherwise) that both run through this code, leading to really unclear exception output. Is there really no way to get a better exception output than this?

requests_kerberos/kerberos_.py
```diff
@@ -71,6 +90,9 @@ def _negotiate_value(response):
         regex = re.compile('(?:.*,)*\s*Negotiate\s*([^,]*),?', re.I)
         _negotiate_value.regex = regex
 
+    if response is None: # we're doing an initial request- just use an empty negotiate value
```
Lukasa: Really minor style nit, but this comment is long enough that I'd rather see it in a block above the `if` than as an in-line comment.

requests_kerberos/kerberos_.py
```diff
@@ -86,21 +108,21 @@ class HTTPKerberosAuth(AuthBase):
...
+            service="HTTP", delegate=False, force_preemptive=False):
         self.context = {}
         self.mutual_authentication = mutual_authentication
         self.delegate = delegate
         self.pos = None
         self.service = service
+        self.force_preemptive = force_preemptive
 
-    def generate_request_header(self, response):
+    def generate_request_header(self, response, host, is_preemptive=False):
```
Lukasa: What does `is_preemptive` do here?

requests_kerberos/kerberos_.py
```diff
@@ -71,6 +90,9 @@ def _negotiate_value(response):
         regex = re.compile('(?:.*,)*\s*Negotiate\s*([^,]*),?', re.I)
         _negotiate_value.regex = regex
 
+    if response is None: # we're doing an initial request- just use an empty negotiate value
+        return ''
+
```
Lukasa: Let's re-evaluate this a little bit.

The purpose of this short-circuit execution, as far as I can tell, is to ensure that `generate_request_header` can build the pre-emptive auth as needed by calling this with a `None` response. However, you've accidentally (presumably as part of a refactor) left a boolean kwarg in `generate_request_header` that is unused.

I wonder if, rather than put the short-circuit all the way down here (in an arguably unrelated bit of code), the higher level function (`generate_request_header`) shouldn't just skip the the call to this function based on the knowledge it has. That would avoid some overhead (a function call and possibly some regex building), and also compartmentalises the logic a lot better.

It's actually orthogonal to the kwarg (you can still signal this with a `None` response), but it seems like that approach would be better. Do you disagree?

requests_kerberos/kerberos_.py
```diff
@@ -285,6 +309,22 @@ def deregister(self, response):
...
+            host = urlparse(request.url).hostname
+
+            with SniffLog(log) as sl:
+                auth_header = self.generate_request_header(None, host, is_preemptive=True)
+
+                if auth_header is None:
+                    # GSS setup failure, return error and details, if any
+                    raise KerberosExchangeError("Kerberos GSS Exchange failed: %s" % str(sl.stored_logs))
+
+            log.debug("HTTPKerberosAuth: Preemptive Authorization header: {0}".format(auth_header))
```
Lukasa: Is it safe to put this header in the logs?

Lukasa:
Fab, thanks! Left multiple notes inline, but this is good progress. =D

requests_kerberos/kerberos_.py
nitzmahone: Yeah, it'll just require collecting those details directly (probably still using a context manager, just not one that's bound to the logger). I'll take a crack at that.

requests_kerberos/kerberos_.py
nitzmahone: Vestigial artifact of a previous incarnation- I'll nuke it

requests_kerberos/kerberos_.py
nitzmahone: Iffy- the non-preemptive version is already being logged, so I just followed suit. If I remember my kerberos details correctly, the value is encrypted by the sk anyway, but if we nuke this one, we should nuke the other as well.

requests_kerberos/kerberos_.py
nitzmahone: That's similar to the way I originally did it, but when doing stuff like this (for a project I don't "own") I tend to go for minimal code change (for ease of review) rather than making bigger changes to the execution flow. I don't have strong feelings either way, but you make a reasonable argument that I can go back to the original way.

requests_kerberos/kerberos_.py
Lukasa: Eh, if we're doing it already let's just leave it. I have really mixed feelings about it, but...we're probably ok.

requests_kerberos/kerberos_.py
Lukasa: I think I prefer the clarity of doing it at a higher level. =)

nitzmahone:
@Lukasa - thanks for looking at another round, let's see if this is more to your liking... Refactored generate_request_header a bit- now it raises with detail instead of returning null on failure (caller can decide what to do with it). Used base EnvironmentException as a temporary "surrogate" error we can raise for the failure code cases since it has an errno (the handler translates those back into KerberosExchangeError). Also restructured a bit to eliminate repetition that gets worse with error raising. Not sure how we'd want to inject detail into the response in the non-preemptive case, so left that one as-is.

requests_kerberos/exceptions.py
```diff
@@ -10,3 +10,8 @@
 
 class MutualAuthenticationError(RequestException):
     """Mutual Authentication Error"""
+
+class KerberosExchangeError(RequestException):
+    """Kerberos Exchange Failed Error"""
+    def __init__(self, *args, **kwargs):
+        super(KerberosExchangeError, self).__init__(*args, **kwargs)
```
Lukasa: We can just leave the class body empty here. =)

requests_kerberos/kerberos_.py
```diff
@@ -86,70 +84,71 @@ class HTTPKerberosAuth(AuthBase):
...
         except kerberos.GSSError as error:
             log.exception(
-                "generate_request_header(): authGSSClientResponse() failed:")
+                "generate_request_header(): {0} failed:".format(kerb_stage))
             log.exception(error)
-            return None
+            raise KerberosExchangeError("%s failed: %s" % (kerb_stage, str(error.args)))
 
-        return "Negotiate {0}".format(gss_response)
+        except EnvironmentError as error:
```
Lukasa: If we aren't using the exception here, don't use the `as` clause. =)

requests_kerberos/kerberos_.py
```diff
@@ -285,6 +284,21 @@ def deregister(self, response):
...
+        if self.force_preemptive:
+            # add Authorization header before we receive a 401
+            # by the 401 handler
+            host = urlparse(request.url).hostname
+
+            auth_header = self.generate_request_header(None, host, is_preemptive=True)
+
+            if auth_header is None:
+                # GSS setup failure, return error and details, if any
+                raise KerberosExchangeError("Kerberos GSS Exchange failed: %s" % str(sl.stored_logs))
```
Lukasa: There are no early returns in `generate_request_header` anymore, so I don't think we can hit this.

Lukasa:
Cool, that's some definite progress! I've made some more notes. =)

requests_kerberos/exceptions.py
nitzmahone: Urg, gotta stop doing this past 1am- more vestigial crap from a different version of that error I missed removing.

requests_kerberos/kerberos_.py
nitzmahone: good call, gone

requests_kerberos/kerberos_.py
nitzmahone: This was another, "worked as written but not as intended"- had intended to ensure the EnvironmentException matched the return code, then translate it to KerberosExchangeError if so (or re-raise if not). Added that (though we could just as easily use a custom internal error type)- the  "translate return codes into an exception" stuff is always messy. Updated.

nitzmahone:
@Lukasa - feedback implemented, ready for another look (hopefully sans dumb late-night issues this time)

Lukasa:
Ok, I think this looks good to me. :sparkles:

@rbcarson, do you want to take a look?

nitzmahone:
@Lukasa - I have one other small unrelated change that Ansible needs (passing a non-default principal to the auth plugin init), but it'd be a conflict with this until it's merged (or I can include in this PR). Assuming those get merged, what's the chances of getting releases cut for this and requests_ntlm?

requests_kerberos/kerberos_.py
```diff
@@ -86,70 +84,75 @@ class HTTPKerberosAuth(AuthBase):
     object."""
     def __init__(
             self, mutual_authentication=REQUIRED,
-            service="HTTP", delegate=False):
+            service="HTTP", delegate=False, force_preemptive=False):
```
rbcarson: This new public setting is worth explaining in the readme file. Setting this to true is what triggers the desired always-reauthenticate-per-request behaviour, correct? Is it sensible to disable the new behaviour by default?

Lukasa:
@nitzmahone I'm happy to do new releases. Feel free to open a new PR once we've got this one merged. =)

rbcarson:
@nitzmahone Is there an easy way to reproduce the problem solved by this PR? I have access to Windows Server, SharePoint, etc.

requests_kerberos/kerberos_.py
nitzmahone: @rbcarson I suppose it depends on how RFC-pedantic you want to be- from [7235](https://tools.ietf.org/html/rfc7235#section-3.1): 
```The "Authorization" header field allows a user agent to authenticate itself with an origin server -- usually, but not necessarily, after receiving a 401 (Unauthorized) response.``` 
I'm not sure what other use cases people use this auth package for, but I know in my case, preemptive auth is a huge perf win.

Comment (nitzmahone) [inline requests_kerberos/kerberos_.py]: I did push a README update describing the new functionality, as well.

**Comment by nitzmahone** [comment]
@rbcarson Sure, for some value of "easy"... ;) You'll need a Windows Server 2008R2+ host that is domain-joined (and that you have full admin rights for), and a *nix-y box that can access port 5985/5986 on the Windows host, and DNS/Kerberos ports on its AD server. The *nix host needs kerberos tools installed, the kerberos realm and DNS configured to point at the AD, and you need to be able to successfully kinit for a user that's in the target server's Administrator group.

Do you have all that? If so, I can whip up some standalone code for a repro that will run against that...

File: requests_kerberos/kerberos_.py (hunk 10 above)
Comment (rbcarson) [inline requests_kerberos/kerberos_.py]: I don't know how RFC-pedantic I am, but the new doc makes the use case clear. :+1:

**Comment by rbcarson** [comment]
@nitzmahone Sounds fairly daunting, but I've been meaning to set up an isolated environment to test this Kerberos stuff - I've only had experience as a connecting client, not as a Kerberos or AD admin. I only have my domain-joined Windows 10 on hand, so I'd appreciate any tips to reproduce this environment, for future testing and debugging purposes.

This PR looks good to me! :+1: Back to you, @Lukasa.

**Comment by nitzmahone** [comment]
@rbcarson If only there were some kind of automation tool around, preferably named for a science-fictiony communication device, that made such things easy... ;)

I actually have an internal sample that stands up an entire AD infrastructure from scratch that we built for a customer - there are several modules that I need to polish up and ship (in my copious free time)... Once those are there, it'd literally be a pushbutton deployment.

**Comment by sigmavirus24** [comment]
@nitzmahone yeah, I'd probably name it after a communication device in science fiction. Probably something starting with an a to be near the start of the alphabet, but also maybe tricorder :-P

File: requests_kerberos/kerberos_.py [hunk 11]
```diff
@@ -86,70 +84,75 @@ class HTTPKerberosAuth(AuthBase):
...
         self.context = {}
         self.mutual_authentication = mutual_authentication
         self.delegate = delegate
         self.pos = None
         self.service = service
+        self.force_preemptive = force_preemptive
 
-    def generate_request_header(self, response):
+    def generate_request_header(self, response, host, is_preemptive=False):
```
sigmavirus24: If we're storing `force_preemptive` does this really need a parameter to the method? Couldn't the method just reference that attribute on the class?

nitzmahone: Only if we want to kill the 401-based thing off. I'm not familiar enough with real-world Kerberos over HTTP to know, but the RFCs mention the possibility for multiple 401/Negotiate cycles before success. If we nuked this and just used the global setting, a (hypothetical) 401 resulting from a preemptive auth request would ignore the previous response and start a new exchange (probably not what we want). But I also have never seen a server that took more than one, so I don't know if that case even works here.

sigmavirus24: Right, that would affect `handle_response` if we relied on the instance variable, right? Yeah, scratch that feedback. *searches for more coffee*

nitzmahone: Hypothetically, yes... 100% of my Kerberos experience has been in Windows bastard-kerberos-land, though, so I have no idea how kerberos functions in the rest of the world. ;)

Lukasa:
Cool, this is good. Thanks so much @nitzmahone! :sparkles: :cake: :sparkles: Feel free to open your other PR now.

PR #70, opened 2016-04-20
add support for principal and realm override

nitzmahone:
Requests-kerberos currently only supports the default principal (ie, whoever the last user you ran kinit/switch for). pykerberos *mostly* supports principal override (it's been broken for awhile, but I'm working on fixing it)- this allows for collection credential caches (like dir and API) to be searched for a matching principal name, and the use of the tickets/TGT contained within. This is a really important feature for Ansible, as we need the ability to support multiple principals simultaneously (kswitch can't do that).

This PR also adds hostname override, a feature that pywinrm needs for parity with its old urllib2-based version. This decouples the hostname in the URL from the kerberos exchange, allowing a host to be accessed via an arbitrary URL hostname (eg for cases of name-based virtual hosting, DNS redirection, hosts files) while its real hostname is used for the kerberos exchange. There are other ways to accomplish this (kerb config, kerb redirection, DNS _kerberos containers, etc), but pywinrm already supports the feature, and I'd rather not have to resort to nasty patches/hacks to keep it working there.

Added tests and docs to README for both features.

### DIFF
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -84,13 +84,16 @@ class HTTPKerberosAuth(AuthBase):
     object."""
     def __init__(
             self, mutual_authentication=REQUIRED,
-            service="HTTP", delegate=False, force_preemptive=False):
+            service="HTTP", delegate=False, force_preemptive=False,
+            principal=None, hostname_override=None):
         self.context = {}
         self.mutual_authentication = mutual_authentication
         self.delegate = delegate
         self.pos = None
         self.service = service
         self.force_preemptive = force_preemptive
+        self.principal = principal
+        self.hostname_override = hostname_override
 
     def generate_request_header(self, response, host, is_preemptive=False):
         """
@@ -108,8 +111,15 @@ class HTTPKerberosAuth(AuthBase):
 
         try:
             kerb_stage = "authGSSClientInit()"
-            result, self.context[host] = kerberos.authGSSClientInit(
-                "{0}@{1}".format(self.service, host), gssflags=gssflags)
+            # contexts still need to be stored by host, but hostname_override
+            # allows use of an arbitrary hostname for the kerberos exchange
+            # (eg, in cases of aliased hosts, internal vs external, CNAMEs
+            # w/ name-based HTTP hosting)
+            kerb_host = self.hostname_override if self.hostname_override is not None else host
+            kerb_spn = "{0}@{1}".format(self.service, kerb_host)
+
+            result, self.context[host] = kerberos.authGSSClientInit(kerb_spn,
+                gssflags=gssflags, principal=self.principal)
 
             if result < 1:
                 raise EnvironmentError(result, kerb_stage)
--- a/README.rst
+++ b/README.rst
@@ -65,7 +65,14 @@ authentication, you can do that as well:
 Preemptive Authentication
 -------------------------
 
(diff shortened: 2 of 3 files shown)


requests_kerberos/kerberos_.py
```diff
@@ -108,8 +111,14 @@ def generate_request_header(self, response, host, is_preemptive=False):
...
             kerb_stage = "authGSSClientInit()"
+            # contexts still need to be stored by host, but realm_override
+            # allows use of an arbitrary kerberos realm (eg, in cases of
+            # aliased hosts, internal vs external, CNAMEs w/ name-based
+            # HTTP hosting)
             result, self.context[host] = kerberos.authGSSClientInit(
-                "{0}@{1}".format(self.service, host), gssflags=gssflags)
+                "{0}@{1}".format(self.service,
+                     host if not self.realm_override else self.realm_override),
+                gssflags=gssflags, principal=self.principal)
```
Lukasa: This is getting really dense. Can we pull the string formatting out of the call to `authGSSClientInit` so that we can more clearly see what's going on?

Lukasa:
I'm totally happy to accept these changes. =) Go forth and write tests!

nitzmahone:
@Lukasa ready for review!

requests_kerberos/kerberos_.py
```diff
@@ -108,8 +111,14 @@ def generate_request_header(self, response, host, is_preemptive=False):
...
         try:
             kerb_stage = "authGSSClientInit()"
+            # contexts still need to be stored by host, but hostname_override
+            # allows use of an arbitrary hostname for the kerberos exchange
+            # (eg, in cases of aliased hosts, internal vs external, CNAMEs
+            # w/ name-based HTTP hosting)
             result, self.context[host] = kerberos.authGSSClientInit(
-                "{0}@{1}".format(self.service, host), gssflags=gssflags)
+                "{0}@{1}".format(self.service,
+                     host if not self.hostname_override else self.hostname_override),
```
Lukasa: Let's hoist this out. At this point this logical line now stretches three physical lines and involves two function calls and an inline conditional: it's just way too complex. Let's break it down slightly. =)

Lukasa:
Left one small note, but this otherwise looks pretty good!

nitzmahone:
@Lukasa broke down authGSSClientInit() arg setup as requested - should be good to go!

requests_kerberos/kerberos_.py
```diff
@@ -108,8 +111,15 @@ def generate_request_header(self, response, host, is_preemptive=False):
 
         try:
             kerb_stage = "authGSSClientInit()"
-            result, self.context[host] = kerberos.authGSSClientInit(
-                "{0}@{1}".format(self.service, host), gssflags=gssflags)
+            # contexts still need to be stored by host, but hostname_override
+            # allows use of an arbitrary hostname for the kerberos exchange
+            # (eg, in cases of aliased hosts, internal vs external, CNAMEs
+            # w/ name-based HTTP hosting)
+            kerb_host = host if not self.hostname_override else self.hostname_override
```
Lukasa: Firstly, let's swap this conditional. Secondly, mind changing this to be `if not None`?

Lukasa:
Alrighty, one small note and then we should be good. I'm happy to prepare some releases. =)

nitzmahone:
Note addressed. That one's always tricky- I usually prefer to have the "defaultish" case upfront instead of hanging off the end, but it's your world...

So about those releases... ;)

Lukasa:
:sparkles: Yay! Ok, I'll aim to release early this week (today is crazy busy for me, so I promise nothing on that specific front).

PR #72, opened 2016-05-16
make mutual auth error sanitization optional

nitzmahone:
Mutual auth error sanitization breaks WinRM long-running operation poll timeout, since it's returned as a 500 without a WWW-Authenticate header. We need a way to allow returning the full error response, as this isn't *really* an error, but we have to see the response to know for sure.

added test coverage, doc update

### DIFF
--- a/test_requests_kerberos.py
+++ b/test_requests_kerberos.py
@@ -403,6 +403,7 @@ class KerberosTestCase(unittest.TestCase):
 
             r = auth.handle_response(response_500)
 
+            self.assertIsInstance(r, requests_kerberos.kerberos_.SanitizedResponse)
             self.assertNotEqual(r, response_500)
             self.assertNotEqual(r.headers, response_500.headers)
             self.assertEqual(r.status_code, response_500.status_code)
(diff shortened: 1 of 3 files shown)


rbcarson:
I'm not very familiar with the internals of Kerberos - would you have any insight as to why `SanitizedResponse` exists? Does the specification say that the server response should be stripped of content and headers when mutual authentication fails?

nitzmahone:
@rbcarson I was wondering the same- I suspect it's more about "you can't necessarily trust this response because we couldn't verify the server's identity", but in this case, we *have* to see the response content...

test_requests_kerberos.py
```diff
@@ -416,6 +417,14 @@ def test_handle_response_500_mutual_auth_required_failure(self):
 
             self.assertFalse(clientStep_error.called)
 
+            # re-test with error response sanitizing disabled
+            auth = requests_kerberos.HTTPKerberosAuth(sanitize_mutual_error_response=False)
+            auth.context = {"www.example.org": "CTX"}
+
+            r = auth.handle_response(response_500)
+
+            self.assertNotIsInstance(r, requests_kerberos.kerberos_.SanitizedResponse)
```
rbcarson: I guess use `assertFalse(isinstance(r, requests_kerberos.kerberos_.SanitizedResponse))` for Python 2.6 support.

nitzmahone: Ah bugger, forgot to check that one

rbcarson:
I'll let this sit for awhile in case @Lukasa or @mkomitee have any comments.

mkomitee:
This is a good idea and you're right about my reasoning. Since this keeps untrusted response sanitizing the default, I'm +1 on this.

mkomitee:
Didn't mean to close.

mkomitee:
Please avoid line continuations. Use tuples instead.

nitzmahone:
Switched line continuation for parenthesized expression.

Lukasa:
This seems reasonable to me. Assuming @mkomitee and @rbcarson have no further notes, either of them may merge.

rbcarson:
:boom:

Thanks, @nitzmahone!

PR #73, opened 2016-05-17
Block use of `principal` argument when using kerberos-sspi. Fixes #71.

### LINKED ISSUE #71: TypeError: authGSSClientInit() got an unexpected keyword argument 'principal'
Looks like now that kerberos-sspi (https://github.com/may-day/kerberos-sspi/blob/master/kerberos_sspi.py#L170) doesn't support 'principal' that was added in commit https://github.com/requests/requests-kerberos/commit/c464cc0035a5552ab28b9e8c3eee7a5904f33996.

rbcarson:
The goal here is to avoid passing the `principal` keyword argument that's unrecognized by kerberos-sspi, while gracefully failing if someone tries to use it with kerberos-sspi / Windows.

This is a holdover until I get around to trying out WinKerberos as the successor to kerberos-sspi + pywin32.

### DIFF
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -122,12 +122,12 @@ class HTTPKerberosAuth(AuthBase):
             kerb_host = self.hostname_override if self.hostname_override is not None else host
             kerb_spn = "{0}@{1}".format(self.service, kerb_host)
             
-            kwargs = {'principal': self.principal}
+            kwargs = {}
             # kerberos-sspi: Never pass principal. Raise if user tries to specify one.
-            if self._using_kerberos_sspi:
-                kwargs = {}
-                if self.principal:
-                    raise NotImplementedError("Can't use 'principal' argument with kerberos-sspi.")
+            if not self._using_kerberos_sspi:
+                kwargs['principal'] = self.principal
+            elif self.principal:
+                raise NotImplementedError("Can't use 'principal' argument with kerberos-sspi.")
 
             result, self.context[host] = kerberos.authGSSClientInit(kerb_spn,
                 gssflags=gssflags, **kwargs)
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -1,7 +1,9 @@
 try:
     import kerberos
+    using_kerberos_sspi = False
 except ImportError:
     import kerberos_sspi as kerberos
+    using_kerberos_sspi = True
 import re
 import logging
 
@@ -94,6 +96,7 @@ class HTTPKerberosAuth(AuthBase):
         self.force_preemptive = force_preemptive
         self.principal = principal
         self.hostname_override = hostname_override
+        self._using_kerberos_sspi = using_kerberos_sspi
 
     def generate_request_header(self, response, host, is_preemptive=False):
         """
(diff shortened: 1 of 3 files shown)


rbcarson:
@nitzmahone This touches your recent new stuff.

@Lukasa Please cut a new release for us Windows users. :)

requests_kerberos/kerberos_.py
```diff
@@ -117,9 +120,13 @@ def generate_request_header(self, response, host, is_preemptive=False):
             # w/ name-based HTTP hosting)
             kerb_host = self.hostname_override if self.hostname_override is not None else host
             kerb_spn = "{0}@{1}".format(self.service, kerb_host)
+            
+            kwargs = {'principal': self.principal}
+            if self._using_kerberos_sspi and self.principal:
+                raise NotImplementedError("Can't use 'principal' argument with kerberos-sspi.")
```
Lukasa: Just to confirm here: does kerberos SSPI reject only truthy values for `principal`, or does it reject it entirely? I want to make sure this fix is in the right direction.

rbcarson: kerberos-sspi doesn't recognize the `principal` keyword argument at all, so it can't even take `None`.

Lukasa: So in that case I think this fix is broken. If `self.principal = None` the code won't enter this block and so will still pass `principal` to the function.

rbcarson:
Okay, all fixed, I think. Caused a shake-up in the unit tests on Windows.

requests_kerberos/kerberos_.py
```diff
@@ -118,9 +121,16 @@ def generate_request_header(self, response, host, is_preemptive=False):
             # w/ name-based HTTP hosting)
             kerb_host = self.hostname_override if self.hostname_override is not None else host
             kerb_spn = "{0}@{1}".format(self.service, kerb_host)
+            
+            kwargs = {'principal': self.principal}
+            # kerberos-sspi: Never pass principal. Raise if user tries to specify one.
+            if self._using_kerberos_sspi:
+                kwargs = {}
+                if self.principal:
+                    raise NotImplementedError("Can't use 'principal' argument with kerberos-sspi.")
```
Lukasa: Ok, the logic here is getting a bit torturous. Maybe initialize `kwargs` to empty dict, then do if/else for the two cases (adding principal to the dict, raising `NotImplementedError`).

Lukasa:
Cool, one note about structure.

nitzmahone:
Checks out OK on Linux/Mac side...

Lukasa:
Beautiful, looks good to me. I'll aim to ship a release with this fix tomorrow. Thanks all!

nitzmahone:
@Lukasa is the 0.10.0 release on pypi hidden? I can see it if I go directly to it, but "latest release" still shows as 0.9.0, and I'm unable to see anything but 0.9.0 when attempting to reference it from extras_require requests-kerberos>=0.10.0.

Lukasa:
Nope, not hidden. @dstufft, any ideas?

nitzmahone:
Hmm, maybe a pip/pypi semver bug from going to 0.10.0 or something. Double check me (but this exact process worked great for 0.9.0) - from a clean venv, do:
```
pip install -i https://testpypi.python.org/pypi --extra-index-url https://pypi.python.org/pypi pywinrm[kerberos]==0.2rc4
```

For 0.10.0, I get 
```
Could not find a version that satisfies the requirement requests_kerberos>=0.10.0 (from pywinrm[kerberos]==0.2rc4) (from versions: 0.9.0)
No matching distribution found for requests_kerberos>=0.10.0 (from pywinrm[kerberos]==0.2rc4)
```

Direct installation of 0.10.0, >=0.10.0, and >0.9.0 all work fine... Very strange.

I don't think it's an issue with requests_kerberos vs requests-kerberos (again, worked fine for 0.9.0), and seems odd that pypi still lists 0.9.0 as latest (maybe that's a semver bug?).

dstufft:
You want ``pypi.python.org/simple/`` and ``testpypi.python.org/simple/`` not ``/pypi``.

nitzmahone:
@dstufft weird that it worked fine for 0.9.0 though- regardless, using /simple *does* fix it. Thanks!

rbcarson:
https://pypi.python.org/pypi/requests-kerberos still shows 0.9.0 as the latest. Maybe there's a stale cache or something.

PR #75, opened 2016-05-29
Replace kerberos-sspi with WinKerberos. Fixes #67.

### LINKED ISSUE #67: Add optional support for WinKerberos
[WinKerberos](https://pypi.python.org/pypi/winkerberos) is a kerberos/pykerberos/kerberos-sspi work-alike for Windows SSPI with no dependencies (no need to install pywin32). It is actively maintained, responsive to bug reports, enhancement requests, and pull requests (though I haven't seen any PRs yet). It supports some features kerberos-sspi does not (the ability to authenticate as a different kerberos user than the process owner) and some features [kerberos-sspi can not](https://github.com/may-day/kerberos-sspi/issues/1). it should be a drop in alternative.

rbcarson:
This reverts the temporary workarounds in #73, and swaps the use of kerberos-sspi with WinKerberos. It's not ready to merge, but looks promising.

@behackett The unit tests fail because WinKerberos doesn't define GSSError, a subclass of KrbError.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -131,10 +131,6 @@ builds). An explicit principal can be specified with the ``principal`` arg:
     >>> kerberos_auth = HTTPKerberosAuth(principal="user@REALM")
     >>> r = requests.get("http://example.org", auth=kerberos_auth)
     ...
-    
-**Windows users:** Explicit Principal is currently not supported when using 
-``kerberos-sspi``. Providing a value for ``principal`` in this scenario will raise
-``NotImplementedError``.
 
 Logging
 -------
--- a/README.rst
+++ b/README.rst
@@ -121,7 +121,7 @@ whom you last ran ``kinit`` or ``kswitch``, or an SSO credential if
 applicable). However, an explicit principal can be specified, which will
 cause Kerberos to look for a matching credential cache for the named user.
 This feature depends on OS support for collection-type credential caches,
-as well as working principal support in pykerberos (it is broken in many
+as well as working principal support in PyKerberos (it is broken in many
 builds). An explicit principal can be specified with the ``principal`` arg:
 
 .. code-block:: python
@@ -131,6 +131,10 @@ builds). An explicit principal can be specified with the ``principal`` arg:
     >>> kerberos_auth = HTTPKerberosAuth(principal="user@REALM")
     >>> r = requests.get("http://example.org", auth=kerberos_auth)
     ...
+
+On Windows, WinKerberos is used instead of PyKerberos. WinKerberos allows the
+use of arbitrary principals instead of a credential cache. Passwords can be
+specified by following the form ``user@realm:password`` for ``principal``.
 
 Logging
 -------
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -1,9 +1,7 @@
 try:
     import kerberos
-    using_kerberos_sspi = False
 except ImportError:
     import kerberos_sspi as kerberos
-    using_kerberos_sspi = True
 import re
 import logging
 
@@ -97,7 +95,6 @@ class HTTPKerberosAuth(AuthBase):
         self.principal = principal
         self.hostname_override = hostname_override
         self.sanitize_mutual_error_response = sanitize_mutual_error_response
-        self._using_kerberos_sspi = using_kerberos_sspi
 
     def generate_request_header(self, response, host, is_preemptive=False):
         """
@@ -121,16 +118,9 @@ class HTTPKerberosAuth(AuthBase):
             # w/ name-based HTTP hosting)
             kerb_host = self.hostname_override if self.hostname_override is not None else host
             kerb_spn = "{0}@{1}".format(self.service, kerb_host)
-            
-            kwargs = {}
-            # kerberos-sspi: Never pass principal. Raise if user tries to specify one.
-            if not self._using_kerberos_sspi:
-                kwargs['principal'] = self.principal
-            elif self.principal:
-                raise NotImplementedError("Can't use 'principal' argument with kerberos-sspi.")
 
             result, self.context[host] = kerberos.authGSSClientInit(kerb_spn,
-                gssflags=gssflags, **kwargs)
+                gssflags=gssflags, principal=self.principal)
 
             if result < 1:
                 raise EnvironmentError(result, kerb_stage)
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -1,7 +1,7 @@
 try:
     import kerberos
 except ImportError:
-    import kerberos_sspi as kerberos
+    import winkerberos as kerberos
 import re
 import logging
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 requests>=1.1.0
-winkerberos >= 0.4.0; sys.platform == 'win32'
+winkerberos >= 0.5.0; sys.platform == 'win32'
 pykerberos >= 1.1.8, < 2.0.0; sys.platform != 'win32'
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 requests>=1.1.0
-kerberos-sspi >= 0.2; sys.platform == 'win32'
+winkerberos >= 0.3.0; sys.platform == 'win32'
 pykerberos >= 1.1.8, < 2.0.0; sys.platform != 'win32'
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 requests>=1.1.0
-winkerberos >= 0.3.0; sys.platform == 'win32'
+winkerberos >= 0.4.0; sys.platform == 'win32'
 pykerberos >= 1.1.8, < 2.0.0; sys.platform != 'win32'
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ setup(
         'requests>=1.1.0',
     ],
     extras_require={
-        ':sys_platform=="win32"': ['winkerberos>=0.3.0'],
+        ':sys_platform=="win32"': ['winkerberos>=0.4.0'],
         ':sys_platform!="win32"': ['pykerberos>=1.1.8,<2.0.0'],
     },
     test_suite='test_requests_kerberos',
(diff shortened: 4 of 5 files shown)


behackett:
> The unit tests fail because WinKerberos doesn't define GSSError, a subclass of KrbError.

Huh. I completely missed that. Easy enough to add.

behackett:
I've added GSSError in WinKerberos master. I can cut a release with this change or wait a bit and see if you run into any other issues. Your call.

rbcarson:
If it's not too much of a bother, please cut a new release. I'm not currently set up to build.

behackett:
[Done](https://pypi.python.org/pypi/winkerberos/0.4.0).

rbcarson:
Tested on Python 3.5.1 x64 with a domain-joined Windows 10 x64 machine, against SharePoint Server 2012 / IIS 8.

```python
>>> from requests_kerberos import HTTPKerberosAuth
>>> import requests
>>> requests.get('http://bd-sharepoint')
<Response [401]>
>>> requests.get('http://bd-sharepoint', auth=HTTPKerberosAuth())
<Response [200]>
>>> requests.get('http://bd-sharepoint', auth=HTTPKerberosAuth(principal='fake@example'))
<Response [200]>
>>> requests.get('http://bd-sharepoint', auth=HTTPKerberosAuth(hostname_override='fake.local'))
<Response [401]>
```

I'm convinced that it works, but the 200 response for the fake principal override value seems suspicious. Can you confirm whether this is the expected behaviour?

behackett:
The 200 response fake@example is weird. I would expect that to fail in authGSSClientInit (at least the docs for [AquireCredentialsHandle](https://msdn.microsoft.com/en-us/library/windows/desktop/aa374713(v=vs.85).aspx) claim it should). My only guess is principal is being ignored. Let me do some research and testing.

behackett:
I have a theory about why AcquireCredentialsHandle is ignoring principal, but no way to test it (my test VMs aren't joined to a kerberos realm). Would you be willing to test a custom build?

rbcarson:
Yeah, for sure. Point me to a wheel and I'll give it a spin.

behackett:
Try this. PR comments don't like the whl type, just rename from zip.

[winkerberos-0.5.0.dev0-cp35-cp35m-win_amd64.zip](https://github.com/requests/requests-kerberos/files/296639/winkerberos-0.5.0.dev0-cp35-cp35m-win_amd64.zip)

The change here is a bit of a wild guess. Various very old sources claim that the pszPrincipal parameter to AcquireCredentialsHandle is ignored by the Kerberos SSP and that you have to use the pAuthData parameter to get access to non-default credentials (WinKerberos exposes this through the user, domain, and password parameters).

rbcarson:
Thanks - I'll get back to you later. I'm currently on Amtrak and don't have the bandwidth to RDP.

rbcarson:
Behaviour appears unchanged in 0.5.0dev0.

@nitzmahone The doc you wrote for the Explicit Principal feature says it "depends on OS support for collection-type credential caches, as well as working principal support in pykerberos (it is broken in many builds)." Any idea what we're missing here?

behackett:
I've been thinking some more about this. On a *nix machine you login locally, then call kinit to get a TGT. Starting with krb5 1.10 you can take advantage of [credential cache collections](http://k5wiki.kerberos.org/wiki/Projects/Credential_cache_collection_improvements) with the addition of kswitch and improvements to klist to get TGTs for multiple principals. gss_acquire_cred can find the correct credentials in the correct cache for the principal you specify.

On Windows the situation is very different. First, there is no kswitch, and klist only lists existing tickets or tgts. It doesn't allow you to get new tickets. The documentation for pszPrincipal appears to be wrong. How it is supposed to work is a mystery. [Comments here](http://www.kerberos.org/software/samples/gsskrb5/gsskrb5-1011/krb5/krb5cred.c) claim that it is ignored in the Kerberos SSP. [This post](https://groups.google.com/forum/#!topic/microsoft.public.platformsdk.security/Xlo23EMiH9U) from back in 2001 claims it must be NULL.

All that said, you can still use an alternative principal. Just split principal at '@' passing the first part to the user parameter of winkerberos.authGSSClientInit and the second part to the domain parameter. For this to work in all situations you'll likely also have to expose the password parameter. Since my dev VMs aren't joined to a kerberos realm and therefore I can't use single sign on, I've tested this extensively and it works well. See [PyMongo's integration](https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/auth.py#L293-L303) as an example.

nitzmahone:
@behackett is correct about the ticket selection behavior with pykerberos. They took my patch to re-enable automatic credential selection via authGSSClientInit, so it works there as expected as of pykerberos 1.1.11 (so long as your OS kerb config has support for collection-typed ccaches). 

That said, if you're not doing that split behavior internally in winkerberos ,I'm guessing you're not attempting to maintain direct-swap API compatibility with pykerberos (eg, exposing new user/domain kwargs)?

behackett:
Well, I was trying to maintain direct-swap API compatibility with pykerberos by supporting principal as documented for SSPI, as well as adding the additional (optional) arguments for user, domain, and password (these are obviously SSPI specific).

The question is, can Windows SSPI be directly compatible with MIT krb5 when it comes to explicit principal handling? This is seriously under documented, but it seems like the answer is no.

rbcarson:
So the facts seems to be:

* `authGSSClientInit()` in WinKerberos ends up ignoring the `principal` argument.
* `authGSSClientInit()` in WinKerberos accepts `user`, `domain`, and `password`, which aren't recognized by PyKerberos.
* Support for `principal` can be emulated when using WinKerberos by splitting on '@' to get the user and domain arguments. But should requests-kerberos do this, or WinKerberos?
* The password *may* be supplied when using the `principal` argument with WinKerberos.

I feel like the change on the requests-kerberos side is shaping up as recognizing a `principal` value in the form of `user[:password]@domain` on Windows, or alternatively adding a `password` argument that will only be recognized on Windows. Does this sound reasonable?

nitzmahone:
@rbcarson I'd probably lean slightly toward the former (```user:password@domain```), just to keep requests-kerberos' API host agnostic, though that has its own problems (e.g., colons and at-signs in passwords). I don't have the bandwidth to dig into the Windows behavior right now to have a stronger opinion... Ironic, since I'm Ansible's "Windows guy", but our controller doesn't run on Windows, so it's a non-issue for us in any case. I will say I'm not generally a fan of "this arg is ignored on most OSs" API design, though.

behackett:
Though user[:password]@domain seems like a good solution (it avoids adding the attractive nuisance of a parameter that only works on one platform), someone somewhere will have ':' in the username part of their user principal.

rbcarson:
I also prefer the `user:password@domain` solution, but ':' is a valid character for the username?? Perhaps we could say that colons in the username must be escaped when the principal value contains a password...

In any case, I think it makes more sense for the special handling of the principal argument to occur in WinKerberos. requests-kerberos would naively pass the value as-is to PyKerberos or WinKerberos without any OS-specific behaviour, but clearly document the differences. What do you think?

behackett:
> but ':' is a valid character for the username??

It's not [disallowed by the RFC](https://tools.ietf.org/html/rfc2743#page-86). More likely you would see special characters like that in the password.

Let me think a bit about changing WinKerberos. It's pre-1.0 and still marked "Beta" in trove classifiers, so I suppose I can still change the API.

rbcarson:
@behackett So, how do you feel about the hack being in WinKerberos instead of requests-kerberos, or vice versa? I suspect that nobody uses principal overrides on Windows anyway, so this is largely academic.

behackett:
Sorry for the long silence. I was out of the office for a month (paternity leave!), then catching up on everything I missed.

I think we can just do this in WinKerberos. Getting rid of the user, domain, and password options before WinKerberos 1.0 would be good for pykerberos API compatibility, and it looks like Windows provides various C functions that should simplify the implementation.

rbcarson:
Cool, I wait for that before making further changes on the requests-kerberos side. And congratulations!

behackett:
I have a patch for this in code review. The format for principal with password will be "user@realm:password". Both the principal (user@realm) and the password can be percent encoded if either have to support ':'. For example, user%40realm:pass%3Aword.

I'll attach another wheel after code review and coverity scan.

rbcarson:
Why is this intentionally different from the typical form of
user:password@host?

behackett:
Well, it's not user@host, it's a kerberos principal, user@realm. User without realm isn't valid AFAIK. It seems natural (and is easy to implement) tacking ":password" onto the principal.

behackett:
@rbcarson looking at the docs for requests-kerberos, the API for passing an explicit kerberos principal wouldn't appear to imply "user:password@realm" would be preferred over "user@realm:password". The credentials aren't passed in the URL (e.g. "https://user:password@host") so it doesn't appear to matter. Did I miss something?

rbcarson:
No, it's just that I've never seen anything of the form "user@realm:password"
until now, and I'm so used to the URL form. Following the URL form for this
non-URL means it'd be possible to break it down using urlparse, but I
suppose that's not necessary in this situation.

behackett:
I'm still waiting on coverity analysis, but this wheel implements the fixes for principal.

[winkerberos-0.5.0.dev0-cp35-cp35m-win_amd64.zip](https://github.com/requests/requests-kerberos/files/525844/winkerberos-0.5.0.dev0-cp35-cp35m-win_amd64.zip)

As before, rename .zip to .whl.

behackett:
@rbcarson, assuming these changes work for you, I'm going to release WinKerberos 0.5.0 next week.

rbcarson:
I won't be able to test this until next week, but I'm confident with your
change.

behackett:
[WinKerberos 0.5.0 has been released](https://pypi.python.org/pypi/winkerberos/0.5.0). Let me know if you run into any issues.

rbcarson:
Okay, I started up IIS 10 on a domain-connected Windows machine and set it up to require Kerberos. Everything seems to work: principal with password, current user as custom principal without password, and other combinations.

@Lukasa For your consideration. I'd like to make a new release shortly after this gets merged.

Lukasa:
Cool @rbcarson, this looks good to me. Thanks so much! :sparkles:

PR #78, opened 2016-07-01
include unit tests in source tarballs

danc86:


Lukasa:
Generally speaking I'm disinclined to do this: can I get an idea of why you want this?

sigmavirus24:
@Lukasa this is typically a request from downstream redistributors. It allows them to use the tarballs on PyPI (instead of retrieving them from somewhere like GitHub) in order to test what they're packaging for end users. It bloats the sdist size, yes, but with wheels, that shouldn't be a terribly big concern anymore.

sigmavirus24:
Also keep in mind that `MANIFEST.in` doesn't affect what gets installed, just what is included in the `.tar.gz` produced by `python setup.py sdist`.

danc86:
Yes this is so that I can run the tests as part of the RPM build for the Fedora package.

Lukasa:
Is there any reason not to use the tarballs from the GitHub tags?

danc86:
I can use Github if you'd prefer.

I would suggest that, in general, source tarballs on PyPI should contain the actual source (matching the sources on Github) -- otherwise they're not really source tarballs are they? -- but it's up to you of course.

danc86:
BTW there is a setuptools plugins [setuptools-git](https://pypi.python.org/pypi/setuptools-git) which can just grab the sources straight from git, so you don't have to enumerate them all in `MANIFEST.in`. But most projects seem to prefer just using `MANIFEST.in`, I guess because it avoids the need for that extra setuptools plugin being installed.

Lukasa:
See, I've never really understood that argument. The PyPI source tarballs are source *distributions*: they're a method of distributing software to users. There is no reason to ship tests to my users, *especially* if I don't install them. I can see some value in doing what, e.g. Twisted does, which involves shipping and installing their tests and test runner such that after you run `pip install Twisted` you can run `trial Twisted`. However, I think most people would argue that that's overkill: I'm just going one step further and not shipping anything in the tarballs that users won't actually use.

danc86:
I guess you are making a distinction between sources need to *build* the package vs. sources need to *develop* the package.

Personally I would suggest there should be no distinction, they should always be the same thing. That lets downstream distributors do things like: run the tests during package build, and ship the complete sources (that is, everything needed to actually *develop* the package, not just build it in) in the source RPM.

Anyway I don't want to waste any more of your valuable time on this :-) so feel free to close this and I can grab the sources from Github instead.

Lukasa:
So that distinction starts getting interesting fast. For example, should I ship .gitignore files? What about, for example, .idea files? The .travis.yml? At a certain point we end up distributing a whole bunch of data in a distribution channel that does not, by and large, care about them.

I see some value in producing an unadulterated tarball of the source tree for downstream repackagers, of course I do. But I don't think that the Python sdist is a good match for that use-case: it's too overloaded, and the effect of making that change is to force the PyPI CDN to distribute a huge amount of extra unnecessary data to accommodate this use-case that is used by *at best* thirty individuals.

PR #89, opened 2017-02-14
Support of SPNEGO authentication

veklov:
Hello,

I have added support of SPNEGO authentication  (in addition to bare Kerberos)
Switched to PyKerberos 1.1.14 and WinKerberos 0.6.0 - both now support corresponding parameters kerberos.GSS_MECH_OID_KRB5 and kerberos.GSS_MECH_OID_SPNEGO.

Could you review and merge the change?

Kind regards,
Alexey Veklov

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,5 @@ env/
 build/
 dist/
 requests_kerberos.egg-info/
+.idea
 
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,4 @@ env/
 build/
 dist/
 requests_kerberos.egg-info/
-.idea
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,7 +1,7 @@
 History
 =======
 
-0.12.0: 2017-01-19
+0.12.0: 2017-02-14
 ------------------
 
 - Switched to PyKerberos 1.1.14 and WinKerberos 0.6.0, implemented support for SPNEGO authentication
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,6 +1,12 @@
 History
 =======
 
+0.12.0: 2017-01-19
+------------------
+
+- Switched to WinKerberos 0.6.0 and implemented support for SPNEGO authentication
+  (just an additional parameter to a kerberos/winkerberos function call).
+
 0.11.0: 2016-11-02
 ------------------
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,8 +4,8 @@ History
 0.12.0: 2017-01-19
 ------------------
 
-- Switched to WinKerberos 0.6.0 and implemented support for SPNEGO authentication
-  (just an additional parameter to a kerberos/winkerberos function call).
+- Switched to PyKerberos 1.1.14 and WinKerberos 0.6.0, implemented support for SPNEGO authentication
+  (just an additional parameter to a pykerberos/winkerberos function call).
 
 0.11.0: 2016-11-02
 ------------------
--- a/README.rst
+++ b/README.rst
@@ -1,8 +1,8 @@
-requests Kerberos/GSSAPI authentication library
+requests Kerberos/Spnego authentication library
 ===============================================
 
 Requests is an HTTP library, written in Python, for human beings. This library
-adds optional Kerberos/GSSAPI authentication support and supports mutual
+adds optional Kerberos/Spnego (GSSAPI - Linux/SSPI - Windows) authentication support and supports mutual
 authentication. Basic GET usage:
 
 
@@ -11,6 +11,11 @@ authentication. Basic GET usage:
     >>> import requests
     >>> from requests_kerberos import HTTPKerberosAuth
     >>> r = requests.get("http://example.org", auth=HTTPKerberosAuth())
+    ...
+
+    >>> import requests
+    >>> from requests_kerberos import HTTPSpnegoAuth
+    >>> r = requests.get("http://example.org", auth=HTTPSpnegoAuth())
     ...
 
 The entire ``requests.api`` should be supported.
--- a/README.rst
+++ b/README.rst
@@ -1,8 +1,8 @@
-requests Kerberos/Spnego authentication library
+requests Kerberos/SPNEGO authentication library
 ===============================================
 
 Requests is an HTTP library, written in Python, for human beings. This library
-adds optional Kerberos/Spnego (GSSAPI - Linux/SSPI - Windows) authentication support and supports mutual
+adds optional Kerberos/SPNEGO (GSSAPI - Linux/SSPI - Windows) authentication support and supports mutual
 authentication. Basic GET usage:
 
 
--- a/requests_kerberos/__init__.py
+++ b/requests_kerberos/__init__.py
@@ -3,7 +3,7 @@ requests Kerberos/GSSAPI authentication library
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Requests is an HTTP library, written in Python, for human beings. This library
-adds optional Kerberos/Spnego (GSSAPI - Linux/SSPI - Windows) authentication support and supports mutual
+adds optional Kerberos/SPNEGO (GSSAPI - Linux/SSPI - Windows) authentication support and supports mutual
 authentication. Basic GET usage:
 
     >>> import requests
--- a/requests_kerberos/__init__.py
+++ b/requests_kerberos/__init__.py
@@ -3,23 +3,27 @@ requests Kerberos/GSSAPI authentication library
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Requests is an HTTP library, written in Python, for human beings. This library
-adds optional Kerberos/GSSAPI authentication support and supports mutual
+adds optional Kerberos/Spnego (GSSAPI - Linux/SSPI - Windows) authentication support and supports mutual
 authentication. Basic GET usage:
 
     >>> import requests
     >>> from requests_kerberos import HTTPKerberosAuth
     >>> r = requests.get("http://example.org", auth=HTTPKerberosAuth())
 
+    >>> import requests
+    >>> from requests_kerberos import HTTPSpnegoAuth
+    >>> r = requests.get("http://example.org", auth=HTTPSpnegoAuth())
+
 The entire `requests.api` should be supported.
 """
 import logging
 
-from .kerberos_ import HTTPKerberosAuth, REQUIRED, OPTIONAL, DISABLED
+from .kerberos_ import HTTPKerberosAuth, HTTPSpnegoAuth, REQUIRED, OPTIONAL, DISABLED
 from .exceptions import MutualAuthenticationError
 from .compat import NullHandler
 
 logging.getLogger(__name__).addHandler(NullHandler())
 
-__all__ = ('HTTPKerberosAuth', 'MutualAuthenticationError', 'REQUIRED',
+__all__ = ('HTTPKerberosAuth', 'HTTPSpnegoAuth', 'MutualAuthenticationError', 'REQUIRED',
            'OPTIONAL', 'DISABLED')
-__version__ = '0.11.0'
+__version__ = '0.12.0'
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -334,7 +334,7 @@ class HTTPKerberosAuth(HTTPGSSAPIAuthBase):
                                     sanitize_mutual_error_response)
 
 class HTTPSpnegoAuth(HTTPGSSAPIAuthBase):
-    """Attaches HTTP GSSAPI/Spnego Authentication to the given Request
+    """Attaches HTTP GSSAPI/SPNEGO Authentication to the given Request
     object."""
     def __init__(
             self, mutual_authentication=REQUIRED,
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -79,14 +79,14 @@ def _negotiate_value(response):
     return None
 
 
-class HTTPKerberosAuth(AuthBase):
+class HTTPGSSAPIAuthBase(AuthBase):
     """Attaches HTTP GSSAPI/Kerberos Authentication to the given Request
     object."""
     def __init__(
-            self, mutual_authentication=REQUIRED,
-            service="HTTP", delegate=False, force_preemptive=False,
-            principal=None, hostname_override=None, sanitize_mutual_error_response=True):
+            self, mech_oid, mutual_authentication, service, delegate, force_preemptive,
+            principal, hostname_override, sanitize_mutual_error_response):
         self.context = {}
+        self.mech_oid = mech_oid
         self.mutual_authentication = mutual_authentication
         self.delegate = delegate
         self.pos = None
@@ -120,7 +120,7 @@ class HTTPKerberosAuth(AuthBase):
             kerb_spn = "{0}@{1}".format(self.service, kerb_host)
 
             result, self.context[host] = kerberos.authGSSClientInit(kerb_spn,
-                gssflags=gssflags, principal=self.principal)
+                gssflags=gssflags, principal=self.principal, mech_oid=self.mech_oid)
 
             if result < 1:
                 raise EnvironmentError(result, kerb_stage)
@@ -321,3 +321,25 @@ class HTTPKerberosAuth(AuthBase):
             # None.
             self.pos = None
         return request
+
+class HTTPKerberosAuth(HTTPGSSAPIAuthBase):
+    """Attaches HTTP GSSAPI/Kerberos Authentication to the given Request
+    object."""
+    def __init__(
+            self, mutual_authentication=REQUIRED,
+            service="HTTP", delegate=False, force_preemptive=False,
+            principal=None, hostname_override=None, sanitize_mutual_error_response=True):
+        HTTPGSSAPIAuthBase.__init__(self, kerberos.GSS_MECH_OID_KRB5, mutual_authentication,
+                                    service, delegate, force_preemptive, principal, hostname_override,
+                                    sanitize_mutual_error_response)
+
+class HTTPSpnegoAuth(HTTPGSSAPIAuthBase):
+    """Attaches HTTP GSSAPI/Spnego Authentication to the given Request
+    object."""
+    def __init__(
+            self, mutual_authentication=REQUIRED,
+            service="HTTP", delegate=False, force_preemptive=False,
+            principal=None, hostname_override=None, sanitize_mutual_error_response=True):
+        HTTPGSSAPIAuthBase.__init__(self, kerberos.GSS_MECH_OID_SPNEGO, mutual_authentication,
+                                    service, delegate, force_preemptive, principal, hostname_override,
+                                    sanitize_mutual_error_response)
--- a/test_requests_kerberos.py
+++ b/test_requests_kerberos.py
@@ -120,6 +120,7 @@ class KerberosTestCase(unittest.TestCase):
                 gssflags=(
                     kerberos.GSS_C_MUTUAL_FLAG |
                     kerberos.GSS_C_SEQUENCE_FLAG),
+                mech_oid=kerberos.GSS_MECH_OID_KRB5,
                 principal=None)
             clientStep_continue.assert_called_with("CTX", "token")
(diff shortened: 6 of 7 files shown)


rbcarson:
This would be really nice to have - thanks! Can you describe how to set up an environment that exercises this new authentication method?

.gitignore
```diff
@@ -4,4 +4,5 @@ env/
 build/
 dist/
 requests_kerberos.egg-info/
+.idea
```
sigmavirus24: This belongs in your global gitignore, not here.

HISTORY.rst
```diff
@@ -1,6 +1,12 @@
 History
 =======
 
+0.12.0: 2017-01-19
```
sigmavirus24: This is in the past.

.gitignore
veklov: Removed

HISTORY.rst
veklov: I have updated the date. Is this file maintained manually, or is it generated automatically and I should not have touched it at all?

rbcarson: This file is maintained manually, and is usually updated by whoever ends up tagging a release for PyPI.

behackett:
I'd also like to know how to explicitly test SPNEGO. The test I added for WinKerberos is a little weak. I assume it passes because the two sides of the conversation only support Kerberos, so that's what they negotiate. I don't have an environment to test NTLM.

README.rst
```diff
@@ -1,8 +1,8 @@
-requests Kerberos/GSSAPI authentication library
+requests Kerberos/Spnego authentication library
```
behackett: Spnego -> SPNEGO

README.rst
```diff
@@ -1,8 +1,8 @@
...
 ===============================================
 
 Requests is an HTTP library, written in Python, for human beings. This library
-adds optional Kerberos/GSSAPI authentication support and supports mutual
+adds optional Kerberos/Spnego (GSSAPI - Linux/SSPI - Windows) authentication support and supports mutual
```
behackett: Spnego -> SPNEGO

requests_kerberos/__init__.py
```diff
@@ -3,23 +3,27 @@
...
+    >>> import requests
+    >>> from requests_kerberos import HTTPSpnegoAuth
+    >>> r = requests.get("http://example.org", auth=HTTPSpnegoAuth())
+
 The entire `requests.api` should be supported.
 """
 import logging
 
-from .kerberos_ import HTTPKerberosAuth, REQUIRED, OPTIONAL, DISABLED
+from .kerberos_ import HTTPKerberosAuth, HTTPSpnegoAuth, REQUIRED, OPTIONAL, DISABLED
```
behackett: Since SPNEGO is an acronym, should this be HTTPSPNEGOAuth?

test_requests_kerberos.py
```diff
@@ -120,6 +120,7 @@ def test_generate_request_header(self):
                 gssflags=(
```
behackett: There should be some tests for SPNEGO.

veklov:
I tested it against two services. One is Java based and uses code form http://spnego.sourceforge.net/ (by the way, I was eventually able to change it to support both "bare" Kerberos token and SPNEGO token that contains Kerberos one). Another service is protected by RSA ClearTrust (some old version which I did not set up and know very little about). 

If you have an Active Directory which you control (at least you can register SPNs against an account you know password for), I can share a simple java/maven project that can be used to start Jetty server with
- SPNEGO authentication filter
- Simple servlet that sends Kerberos principal back in response

README.rst
veklov: Fixed

README.rst
veklov: Fixed

requests_kerberos/__init__.py
veklov: I have seen Spnego, SPNego, SPNEGO. I personally do not like HTTPSPNEGOAuth as it is difficult to read.

rbcarson:
@veklov I'd appreciate anything you can provide to help me repro this environment. I have access to an Active Directory domain with the ability to register SPNs.

Also, I'd also appreciate it if anyone can explain the scenario that requests-kerberos previously didn't support. I'm not familiar with the details of what's happening here. When a server supports either NTLM or Kerberos auth (or only Kerberos), it will always respond with `www-authenticate: Negotiate`, right? Is deciding on whether to use Kerberos or NTLM the only thing being negotiated? It sounds like no, because adding SPNEGO support doesn't mean requests-kerberos can now handle a negotiation to NTLM.

requests_kerberos/__init__.py
sigmavirus24: I think what @veklov has is fine.

veklov:
@rbcarson Hi,

As far as I understand (I am not a guru in this question), there are three different types of payload for Negotiate:

1) NTLM token - start with TlRMTVNT (it is literally prefix in base64)
This is something ancient. I see this only when I have an issue with my service setup.
If Internet Explorer receives "Negotiate", but it cannot generate Kerberos token (say because SPN is not registered in AD, or DNS name is not properly set up) - it re-tries request sending NTLM token. 

2) Kerberos token - "bare" Kerberos token
I believe it is base64 encoded Kerberos token without any wrappers, prefixes or additions.

3) SPNEGO token - "wrapped" Kerberos token.
Actually SPNEGO token is a container of authentication method ids/tokens. For example, Internet Explorer (I suspect all clients who uses Windows SSPI) sends the following OIDs/tokens in SPNEGO token
	1.2.840.48018.1.2.2 - Legacy OID for Kerberos V5
	1.2.840.113554.1.2.2 - Kerberos V5
	1.3.6.1.4.1.311.2.2.30 - something new and modern :) 
	1.3.6.1.4.1.311.2.2.10 - same :)
And server can chose which OID/token to accept based on what it supports and priorities.

As I mentioned in my previous post, I have tested two services (servers): Java "knows" only 1.2.840.113554.1.2.2 - Kerberos V5, while Clear Trust (at least my version) understands only 1.2.840.48018.1.2.2 - Legacy OID for Kerberos V5.

It looks like MIT Kerberos also did something to support 1.2.840.48018.1.2.2 - Legacy OID for Kerberos V5 (or something else changed in my env). Version 1.6 does not work against Clear Trust, while version 1.14 works. 
 
@behackett "... passes because the two sides of the conversation only support Kerberos, so that's what they negotiate." - I think important fact is that server can extract Kerberos token from SPNEGO one. Even if both sides support only Kerberos as authentication mechanism, they must be able to generate/parse SPNEGO token containing Kerberos one. If your test covers this, than I believe it is fine.

For true interoperability test I will try to "package" my Java/Maven example.

Kind regards,
Alexey Veklov

veklov:
Hi @rbcarson, @behackett,

Attached is the example. See readme.md for details. Requires Java 1.8.
Let me know if something does not work (there may be differences in the env I did not anticipate)

Kind regards,
Alexey Veklov
[spnego.zip](https://github.com/requests/requests-kerberos/files/786087/spnego.zip)

veklov:
Hi @rbcarson, @behackett,

I have done more experiments with the example I provided above (It accepts only SPENGO tokens with Kerberos mechanism. on GSSAPI level it accepts SPNEGO mechanism). Below are 4 scenarios I tested.

- cases 2) and 3) - SPNEGO containing Kerberos or NTLM. Matches specifications I have seen so far. 
- case 1) - what I previously called "bare" Kerberos. I am not sure what it is, and I have not seen specs on this. It would work if I changed server side GSSAPI to accept two mechanisms SPNEGO and Kerberos (or just Kerberos).
- case 4) - According to source from  http://spnego.sourceforge.net/, it is NTLM, but again according to specs it must not be prefixed with Negotiate, but with NTLM, so I am confused.

1) requests_kerberos.HTTPKerberosAuth
HTTP header: Negotiate YIIJMgYJKoZIhvcSAQICAQBu ... OzQJooehaKzuP3sdoJ9vYumCrL
Authentication fails. 
When I try to parse token using org.jaaslounge.decoding.spnego.SpnegoInitToken if also fails with message: Not a valid SPNego token

2) requests_kerberos.HTTPSpnegoAuth + correct DNS + correct SPN
HTTP header: Negotiate YIIJgAYGK ... 45SgXfi
Authentication succeeds:
Token decoder shows
        spnegoToken.getMechanism() = 1.2.840.48018.1.2.2
            spnegoToken.getMechanisms():
                1.2.840.48018.1.2.2
                1.2.840.113554.1.2.2
                1.3.6.1.4.1.311.2.2.30
                1.3.6.1.4.1.311.2.2.10

3) requests_kerberos.HTTPSpnegoAuth + "localhost" as DNS
HTTP header: Negotiate YHoGBisGAQUFAqBwMG ... TMzNURCRw==
Authentication fails
Token decoder shows
            spnegoToken.getMechanism() = 1.3.6.1.4.1.311.2.2.10
            spnegoToken.getMechanisms():
                1.3.6.1.4.1.311.2.2.10
                1.2.840.48018.1.2.2
                1.2.840.113554.1.2.2
                1.3.6.1.4.1.311.2.2.30

4) requests_kerberos.HTTPSpnegoAuth + correct DNS but without corresponding SPN
HTTP header: TlRMTVNTUAABAAAAl4II4gAAAAAAAAAAAAAAAAAAAAAGAbEdAAAADw==
Authentication fails. 
Token decoder fails with message: Malformed SPNego token

veklov:
Hi @rbcarson, @behackett,

To summarize above:

1) Current version of requests_kerberos implements something non-standard. It sends unwrapped Kerberos token. For example, Apache HTTP Client for java does the same, but uses "Kerberos" instead of "Negotiate" in the HTTP header. It looks like fortunately GSSAPI/SSPI can somehow distinguish Kerberos and SPNEGO tokens when you pass content of "Negotiate" to them (tested with Java implementation of GSSAPI by passing both mechanisms to GSSAPI on server side)

2) Some sort of negotiation is possible if your server accepts multiple authentication schemes by sending something like this: WWW-Authenticate: Kerberos, NTLM, Basic realm="REALM NAME" - and you add multiple handlers on client side:
    session.auth = SomeDispatcherBasedOnValueOfWWWAuthenticate(
       requests_kerberos.HTTPKerberosAuth(mutual_authentication=requests_kerberos.OPTIONAL),
       requests_ntlm.HttpNtlmAuth('domain\\username','password'),
       ,,,
   )
At least in Apache HTTP Client this is done this way.

3) Negotiation in terms of https://tools.ietf.org/html/rfc4559 is the goal of this pull request.
But actual authentication mechanisms available for negotiation depend in GSSAPI on Linux, and SSPI on Windows.

test_requests_kerberos.py
veklov: I have added test_passes_correct_mech_oid. Please take a look.

veklov:
@rbcarson, @behackett, @sigmavirus24 Hi,
I think I covered all code review comments. Could you take a look if more changes required?
Have you had a chance to try the example?

behackett:
@veklov I'm gonna have to trust you on the SPNEGO details. Your knowledge clearly far exceeds my own. I'm not a maintainer of this project, just an interested observer.

I wonder if your changes would help in https://github.com/requests/requests-kerberos/issues/91.

rbcarson:
This pull request is on my to-do list - I haven't had the time to try the examples and dive into this issue. Sorry!

wkmanire:
Any idea when this might be merged?

sodi07:
Same comment as @wkmanire  - Any idea when this could be merged?

nitzmahone:
Hrm, I have mixed feelings about relying on a black-box native code NTLM impl, especially if it still requires a different auth handler. If you already don't mind using a different auth handler, why not just use https://github.com/requests/requests-ntlm? It's actively maintained and heavily used by Ansible/pywinrm.

segevfiner:
Aren't you already relaying on a black box native implementation of Kerberos on Windows? Is there really anything wrong with using platform APIs?

The thing is that many servers don't support plain Kerberos. They require negotiate which can resolve to either Kerberos or NTLM. Supporting negotiate and ntlm here will make this work transparently, and I think also allows for Windows SSO even for NTLM if you use SSPI to implement this (Which should already happen for Kerberos if you use SSPI).

I think requests_ntlm is pure Python and as such can't support Windows SSO.

nitzmahone:
Sorry @veklov, I think I misunderstood what you were trying to accomplish here based on a quick skim of the PR and others' comments. It looks like you're only trying to change the MECH_OID to optionally support using SPNEGO (eg for servers that don't accept `GSS_MECH_OID_KRB5`), but not actually implementing the "Negotiate" part of the functionality (eg, it will only work with Kerberos and not allow the NTLM fallback). Is that accurate? If so, I generally don't have a problem with that, although I'm not sure why it's necessary to bump the minimum pykerb version. I also think it's important to make that distinction clearer in the documentation, as I think a casual observer would expect that SPNEGO support includes the NTLM fallback behavior. If not for the possibility of implementing NTLM fallback support at some point in the future (eg, bringing in parts of requests-ntlm), I'd say just ditch the base class altogether and allow an OID override arg on the constructor... But NTLM override (at least on POSIX systems) would require different constructor args, like making the principal name and a password required values, so the two derived implementations set us up well for that split in the future, should it ever happen.

mkomitee:
FWIW, ccs-kerberos now exports GSS_MECH_OID_SPNEGO, and I've verified that it works fine in my environment.

pu239ppy:
This would be really great to have, I find that when making kerberized connection to Jetty I have to explicitly set kerberos.GSS_MECH_OID_SPNEGO as the underlying mech OID.  I also filed a duplicate (albeit a much simpler one, this was before I became aware of this PR) as #115 and would be happy to close it if this one gets merged.

mkomitee:
I've been thinking about this, and I think it makes sense for there to be a SPNEGO authentication handler for requests which transparently handles GSSAPI/Kerberos and NTLM. 

The name of this project is requests-kerberos and naming is important (for things like discoverability) so I'm not sure it makes sense for it to be this one.

That said, there are going to be lots of similarities between the two, so may it makes sense to fork this into github.com/requests/requests-spnego and work on it there?

requests_kerberos/kerberos_.py
```diff
@@ -79,14 +79,14 @@ def _negotiate_value(response):
     return None
 
 
-class HTTPKerberosAuth(AuthBase):
+class HTTPGSSAPIAuthBase(AuthBase):
```
mkomitee: Do we intend for users to wrap this base-class further? If not, maybe we should prefix it with an `_` to make it easier to change in the future without breaking peoples code?

jborean93:
Ultimately I'm trying to reduce the scope of this library and just have it focus on Kerberos. There are a few things already in this library that are Kerberos specific; mutual auth, delegation, etc. By adding SPNEGO we bring in NTLM which is really not within the scope of Kerberos and I recommend you use a separate library.

alexey-veklov-clearscale:
@jborean93 SPNEGO is a negotiation protocol and envelope. It is not NTLM specific and does not bring NTLM in. For example, Linux SPNEGO and Java SPNEGO "implementations" do not support NTLM and always "negotiate" to use Kerberos or fail. In case you talk to a service that supports kerberos, but has no knowledge about nonstandard way of putting bare kerberos token into the authorization header, support for passing Kerberos token over SPNEGO makes the difference.

jborean93:
I understand that but essentially SPNEGO will "negotiate" between Kerberos and NTLM in most common situations. We don't want to handle anything outside of Kerberos and even if we do want to support it we run the risk of bringing in an auth protocol that requires more than 1 round trip to complete the auth which this library does not support. I'm not aware of any GSSAPI implementation that supports saying you want just Kerberos auth over an SPNEGO wrapper.

While yes things may require the SPNEGO wrapper around the Kerberos token I feel that this library is just for pure Kerberos rather than adding in Negotiate. If you need that support then you should look at other libraries which offer such a thing. Adding support for Negotiate is a slippery slope and honestly I'm just trying to clean up this library to use more modern libraries rather than add new functionality into it.

alexey-veklov-clearscale:
I added this request after I successfully used it on client side with Linux GSSAPI (PyKerberos) and Windows SSPI (WinKerberos) to talk to a native Windows and Java SPNEGO servers

jborean93:
You are free to fork it in your own library or use another one. Ultimately I'm just trying to reduce the scope of `requests-kerberos` to focus solely on Kerberos. I know this isn't what people may want to hear but there needs to be a limit in what is offered here. Starting to offer SPNEGO will open up the possibility of exchanging more than just Kerberos tokens which is completely outside of what this library is for.

gpchelkin:
We can all just use https://github.com/pythongssapi/requests-gssapi (for Linux GSSAPI) and https://github.com/brandond/requests-negotiate-sspi (for Windows SSPI) instead.

gpchelkin:
BTW here is how I use it in cross-platform snippet (for the sake of completeness):
```python
import platform
import requests
system = platform.system()

if system in ['Linux', 'Darwin']:
# https://github.com/pythongssapi/requests-gssapi
# https://github.com/pythongssapi/python-gssapi
    import gssapi
    from requests_gssapi import HTTPSPNEGOAuth
    auth = HTTPSPNEGOAuth(mech=gssapi.mechs.Mechanism.from_sasl_name("SPNEGO"))
elif system == 'Windows':
# https://github.com/brandond/requests-negotiate-sspi
    from requests_negotiate_sspi import HttpNegotiateAuth
    auth = HttpNegotiateAuth()
 
r = requests.post("https://example.com/login", auth=auth)
```

abatkin:
In practice, SPNEGO _is_ Kerberos, just with an extra semi-standardized wrapping around the tickets. I don't think anyone wants this library to support NTLM or anything else that might somehow get negotiated. It's just that the SPNEGO-style tickets are much more widely used than "plain" Kebreros.

jborean93:
> In practice, SPNEGO is Kerberos

I wouldn't go that far, SPNEGO is more than just Kerberos as it includes both NTLM and NegoEx from Microsoft with the latter being a more advanced version of SPNEGO (better negotiation of auth protocols). I've yet to see other SPNEGO mechs in GSSAPI but considering they support NegoEx it is possible for it to be done on that platform anyway. You could say that Kerberos is most of the time done through SPNEGO and I would probably agree though. At least on SSPI if the acceptor has requested SPNEGO but the client offers a raw Kerberos token then SSPI will handle that accordingly. Unfortunately the GSSAPI mechanisms will not which is the crux of the problem here.

_A quite side note: I found that SSPI doesn't support the opposite, a Kerberos token wrapped in SPNEGO but the acceptor only allows Kerberos. This is a pretty niche scenario but still is something to keep in mind._

I'm not against having something that implements the SPNEGO wrapping of Kerberos tokens but this would need to ensure that it's just SPNEGO with Kerberos and no other mech. This is a pretty important distinction because without this `requests-kerberos` will now offer more than just Kerberos which means:

* The delegation code may or may not work anymore (not all providers support delegation)
* The mutual auth code may or may not work (not all providers support mutual auth)
* Some auth protocols require more than just 1 exchange of tokens (NTLM requires 2 round trips)

Currently with the main 3 client providers that are used here are the difficulties with restricting what SPNEGO/Negotiate can offer:

* MIT has implemented `gss_set_neg_mechs` in 2010 which which looks like is part of the `1.9` release
  * The `python-gssapi` module that `pyspnego` exposes this function so `pyspnego` could theoretically call this
* Heimdal has only just implemented this in 2019
  * https://github.com/heimdal/heimdal/issues/495
  * The latest release of Heimdal is 7.7.0 and this does not include the change so this won't work for Heimdal (yet)
  * macOS uses a forked copy of Heimdal from even further back and even the latest release from Apple https://opensource.apple.com/source/Heimdal/Heimdal-597.140.2/ does not include this
* SSPI (Windows) has no way of controlling what protocols SPNEGO will provide
  * The closest I could find is (SetCredentialsAttributesW)[https://docs.microsoft.com/en-us/windows/win32/api/sspi/nf-sspi-setcredentialsattributesw) but the `SECPKG_ATTR_SUPPORTED_ALGS` and `SECPKG_ATTR_SUPPORTED_PROTOCOLS` are for SChannel only (not Negotiate)
  * Someone is free to correct me but I assume it's not possible

So ultimately the only way to properly achieve Kerberos only through SPNEGO is on MIT which starts to fragment the supported platforms even more. Luckily all is not lost and it would be possible to achieve this outside of the GSSAPI/SSPI provider through `pyspnego`. The code currently cannot do this today but it would be possible for `pyspnego` to act as the SPNEGO wrapper and it calls GSSAPI/SSPI to get the raw token to wrap. It also has the Python implemented SPNEGO provider that does this but it would need extra work to restrict the mechs it uses to just Kerberos.

I hope that explains the problem as I see it and I welcome attempts to bring this in but first off it would have to be plumbed into `pyspnego` in a cross platform fashion before this library could utilise it.

PR #92, opened 2017-04-10
Added support for passing along the CBT token to the pykerberos library

jborean93:
This PR adds in support for passing along a CBT token to pykerberos for endpoints that require this to be set. It is still pending a review and approval for the required changes https://github.com/02strich/pykerberos/pull/27. Would be good to get your initial impression while I am waiting for that though.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -161,6 +161,22 @@ else
     echo -e "SUCCESS: Apache site built and set for Kerberos auth\nActual Output:\n$CURL_OUTPUT"
 fi
 
+if [ "$IMAGE" == "ubuntu:16.04" ]; then
+    echo "Downloading Python $PYENV"
+    wget -q "https://www.python.org/ftp/python/$PYENV/Python-$PYENV.tgz"
+    tar xzf "Python-$PYENV.tgz"
+    cd "Python-$PYENV"
+
+    echo "Configuring Python install"
+    ./configure &> /dev/null
+
+    echo "Running make install on Python"
+    make install &> /dev/null
+    cd ..
+    rm -rf "Python-$PYENV"
+    rm "Python-$PYENV.tgz"
+fi
+
 echo "Installing Pip"
 wget -q https://bootstrap.pypa.io/get-pip.py
 python$PY_MAJOR get-pip.py
--- a/.travis.sh
+++ b/.travis.sh
@@ -89,6 +89,17 @@ mkdir -p /var/www/example.com/public_html
 chmod -R 755 /var/www
 echo "<html><head><title>Title</title></head><body>body mesage</body></html>" > /var/www/example.com/public_html/index.html
 
+echo "Create self signed certificate for HTTPS endpoint"
+mkdir /etc/apache2/ssl
+openssl req \
+    -x509 \
+    -nodes \
+    -days 365 \
+    -newkey rsa:2048 \
+    -keyout /etc/apache2/ssl/https.key \
+    -out /etc/apache2/ssl/https.crt \
+    -subj "/CN=$KERBEROS_HOSTNAME/o=Testing LTS./C=US"
+
 echo "Create virtual host files"
 cat > /etc/apache2/sites-available/example.com.conf << EOL
 <VirtualHost *:80>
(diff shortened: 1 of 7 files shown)


behackett:
@rbcarson - FYI - it doesn't look like SSPI supports client channel bindings (there is no way to specify this in InitializeSecurityConext, though AcceptSecurityContext on server side supports it), so there is no way to add support to WinKerberos.

https://en.wikipedia.org/wiki/Security_Support_Provider_Interface#Comparison

rbcarson:
What are some examples of servers/services/products that require channel
bindings? Is the implication then that Windows would always need third
party software to authenticate with such endpoints?

jborean93:
@rbcarson The scenarios where CBT is required from my knowledge is when connecting to ISS and extended protection is set [details here])https://www.iis.net/configreference/system.webserver/security/authentication/windowsauthentication/extendedprotection) or WinRM with the CbtHardeningLevel set to Strict [details here](https://msdn.microsoft.com/en-us/library/cc251568.aspx). I'm wanting this through for the latter as my org requires this to be set to strict and it stops us from using Ansible with Kerberos auth

@behackett I just had a quick look this morning because I assumed SSPI did some background magic with CBT automatically but I was wrong and my tests failed from a Windows host. Looking at the docs very briefly at the MS docs it seems like the `pInput` parameter in [InitializeSecurityContext](https://msdn.microsoft.com/en-us/library/windows/desktop/aa375507(v=vs.85).aspx) is where the Channel Bindings needs to be set. This is done through a [SecBuffer](https://msdn.microsoft.com/en-us/library/windows/desktop/aa379814(v=vs.85).aspx) of the type [SEC_CHANNEL_BINDINGS](https://msdn.microsoft.com/en-us/library/windows/desktop/dd919963(v=vs.85).aspx). Depending on how the pykerberos changes go theoretically there could be a similar method `buildChannelBindingsStruct` to return a pointer to the SecBuffer struct of Channel Bindings which is then passed through to `authGSSClientStep` but TBH I'm not sure what they will say about my changes there.

behackett:
pInput has to be NULL on the first call to InitializeSecurityContext. On subsequent calls it is used to pass the server's challenge.

Various sources seem to at least strongly imply that if TLS is in use through sChannel the SSPI stack automatically adds a CBT to the client responses. The only information I could find about passing an explicit CBT is a discussion here (with no resolution):

https://social.msdn.microsoft.com/Forums/windowsdesktop/en-US/3ecb99c5-542e-4bcd-a808-116eb915cff8/enabling-channel-binding-for-kerberos-authentication-in-a-3rd-party-library-openssl?forum=windowssecurity

jborean93:
I am probably wrong on this but from how I read it pInput is NULL on the first call but on subsequent calls it points to a `SecBufferDesc` which in their words is `The SecBufferDesc structure describes an array of SecBuffer structures...` which sounds to me you can have multiple `SecBuffer` structures in that array, ` for the server's challenge and another for the Channel Bindings.

behackett:
Ah! Yes, you are correct. Sorry about that.

jborean93:
Now you've got me interesting in doing the same for winkerberos :), I'll try and get a basic POC sometime soon to prove it works and potentially look at mirroring the eventual changes in pykerberos once I hear back from them.

behackett:
Cool. Patches welcome.

rbcarson:
@jborean93 Do you know the CBT support situation in the Apple Kerberos project that 02strich/pykerberos is forked from?

I set Extended Protection to Required on my domain-connected Windows 10 IIS instance (no SSL), but I was still able to authenticate successfully using requests-kerberos+WinKerberos.

Despite being a maintainer for requests-kerberos, my understanding of the low-level stuff at play here is superficial at best. At this point I have high-level concerns about divergence across:

* WinKerberos
* Apple Kerberos, for which we've had a request from at least one Linux distro maintainer to return to
* @02strich's fork used by requests-kerberos that shares the same `kerberos` namespace as Apple Kerberos

At a glance, this pull request needs to add a guard against using keywords and functions that don't exist on WinKerberos (yet?). Plus unit tests, of course. :)

jborean93:
@rbcarson no worries understand the concern and was hoping these PRs would bring them all to light so we can work through them. css-pykerberos/kerberos currently parses in GSS_C_NO_CHANNEL_BINDINGS to the `authenticate_gss_client_step` method like 02strich/pykerberos does so it would have the same isuess.

There is an open question from @nitzmahone there regarding the status of the fork and whether we should also make the changes there, my thoughts are yes we should but waiting to hear back. I'm planning on looking at winkerberos sometime soon based on the conversation here to add in the same functionality.

Ultimately I'm not sure I have written the code correctly in pykerberos and was hoping for feedback before propagating the changes across the 3 Kerberos libraries. Based on this feedback I am trying to make the implementation the same across all 3 to reduce the complexity required but time will tell if that is possible.

The guards I have put in place really is around checking if the method `buildChannelBindingsStruct` exists in the kerberos library, if it does them it assumes `authGSSClientStep` also takes in the `input_chan_bindings` kwarg. If the build method doesn't exist then it won't try and add the channel bindings to keep compatibility with winkerberos. Happy to revisit this if there are better ways around this.

As for the Windows 10 IIS with Extended Protection I believe the way it works under SSL is different from not having SSL but I can't find anything to verify that at the moment. This is definitely required for WinRM HTTPS with CBT set to strict and I can easily replicate it not working before and working after this fix.

jborean93:
After hearing back from 02strict/pykerberos the intention is to make the changes in css-pykerberos/kerberos instead and I've raised a PR there https://github.com/apple/ccs-pykerberos/pull/55. Will hopefully hear back from them soon but this PR might end up requiring a change to the kerberos dependency.

jborean93:
Hi @behackett @carsonyl 

I think I am finally ready to revisit this PR and see if I can get it merged into devel. The underling changes that are required for this PR have been merged in with;

* pykerberos (awaiting release to pypi) - https://github.com/02strich/pykerberos/pull/27
* winkerberos (released to pypi) - https://github.com/mongodb-labs/winkerberos/pull/17
* ccs-pykerberos (awaiting release to pypi) - https://github.com/apple/ccs-pykerberos/pull/55

The underlying methods that are called are all the same in the 3 libraries and if the decision is made to move back to Apple's fork of pykerberos in the future the functionality is there as well.

If someone has some time to review the changes in this PR, I'm happy to adjust whatever is required so please let me know.

Thanks

PR #103, opened 2017-11-13
[RFC] port from pykerberos to python-gssapi

frozencemetery:
We (Fedora & RHEL/CentOS) have been working toward deprecating pykerberos (and python-krbV) in favor of the more maintainable, easier-to-use [python-gssapi](https://github.com/pythongssapi/python-gssapi).  The current intent is to remove both from Fedora 28.  Of course, we'd like to keep requests-kerberos around, hence: patch.

This is probably not what you want the final product to look like, should you decide to accept this change.  I've taken a number of shortcuts: the most obvious is that I didn't touch the tests (because they're very tied into how pykerberos operates), and I've also removed the SSPI support (we believe python-gssapi should work on Windows, but would require an MIT/Heimdal installation there, and so isn't really helpful).  Should you decide to move forward with this approach, I'm happy to adjust the code/tests to do all this properly; this is just to demo python-gssapi.

The alternative to switching off pykerberos is for us to fork and maintain separately (requests-gssapi?), which I'd rather not do because it's confusing and splits the install base.

What do you think?

### DIFF
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -1,9 +1,8 @@
-try:
-    import kerberos
-except ImportError:
-    import winkerberos as kerberos
 import re
 import logging
+
+import gssapi
+from gssapi.exceptions import GSSError
 
 from requests.auth import AuthBase
 from requests.models import Response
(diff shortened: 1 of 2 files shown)


nitzmahone:
I've been noodling on the same thing for awhile; the hell we just went through to get WinRM-compatible IOV encryption implemented would've been a lot easier on `python-gssapi`. 

I don't really have a problem with switching the main stuff over, but we need to preserve the Windows code path on winkerberos, which will unfortunately make the code a lot more complex...

behackett:
It will make the code more complex since the two libraries have very different APIs. WinKerberos has a few advantages though. The largest, I think, is that binary wheels exist for it since SSPI has a stable ABI and exists on every version of Windows Microsoft still supports. Installing MIT Kerberos, the right VC++ version for your version of CPython, and figuring out how to make the build system find the krb5 headers on Windows is a pretty high bar for end users.

frozencemetery:
If you all want to port over, that's great and I'm happy to answer any python-gssapi questions you may have.  However, hearing nothing in the past few weeks, and seeing the other PRs which are unmerged, I have gone ahead and put work into [requests-gssapi](https://github.com/pythongssapi/requests-gssapi), so I am not personally invested in porting requests-kerberos and do not plan to contribute further code at this time.

PR #104, opened 2017-12-07
add winrm encryption passthru support

nitzmahone:


setup.py
```diff
@@ -52,7 +52,7 @@ def get_version():
     ],
     extras_require={
         ':sys_platform=="win32"': ['winkerberos>=0.5.0'],
-        ':sys_platform!="win32"': ['pykerberos>=1.1.8,<2.0.0'],
+        ':sys_platform!="win32"': ['pykerberos>=1.2.1dev0'],
```
jborean93: Do we want to keep the `dev0` in the requirements once pykerberos is actually released?

nitzmahone: No- this is just so we can tell when someone's running a from-source version of the code. We'll bump it to a proper value during releases and increase the value in master at that point. I'm going to start doing this in the Ansible repo as well when I'm RM...

nitzmahone: (just realized this is the setup.py side) - that's up for debate. We could probably leave it at its previous value here, but bump it for pywinrm[kerberos]...

Issue #97, opened 2017-06-12
Python kerberos library selection 


nitzmahone:
With the `pykerberos` fork effectively going dark (see discussion at https://github.com/02strich/pykerberos/pull/27) and Apple's original ccs-pykerberos being apparently erstwhile maintained, I'm wondering what we can do to be on a supported Kerberos layer that's being actively maintained? pywinrm has a need to merge some specific IOV wrapping code that can only be implemented with low-level access to the C library and structures- we've got the code written, but Apple appears to have gone dark again (even if I could get the changes merged, I have no confidence that we could get any future bugfixes merged/released quickly). One option might be https://github.com/pythongssapi/python-gssapi - it uses Cython to provide access to both a high-level API and the underlying C API, so we could actually reimplement what we have in pure Python if this library were used. Biggest problem I can see there would be the Windows stuff that's mirroring the pykerberos library- we'd have to effectively split the implementation or write an interposer that would mimic the pykerberos API (since AFAIK there's no Windows equivalent available). We could also of course just make yet another fork of the original ccs-pykerberos...

We might (at least temporarily) take the gssapi approach with pywinrm and vendor a hacked requests-kerberos inside it, as that frees us from the reliance on third parties to get bugs fixed and released, but I'm curious if the requests team has any thoughts or different ideas about this problem in general?

nitzmahone:
(kept meaning to hassle @Lukasa at Pycon about this but forgot)

Lukasa:
Honestly, I think we don't have an opinion on this as a team. Generally speaking I think our position on this is that we want to use an actively maintained Kerberos implementation if we can, and we also don't want to be jumping around between implementations. 

So far we haven't managed to do that very well: I'd like to find a way to do better. Sadly, I have no concrete advice as to what to do here: this is just not a space I know much about.

02strich:
@nitzmahone has apple come back on your pull requests regarding the additional IOV features? At this point, what is the level of confidence that apple will keep maintaining upstream?

behackett:
I don't see a pull request for ccs-pykerberos to add the IOV wrapping code. Did they previously reject it?

nitzmahone:
[...]

Issue #63, opened 2016-02-26
Switch to kerberos instead of pykerberos


toabctl:
Distributions (openSUSE, Debian, Ubuntu, Fedora) already ship python-kerberos but nobody ships python-pykerberos . And both have the same kerberos.so so they are not installable in parallel.
Can you please switch to kerberos (from pypi) instead of using pykerberos? Looks anyway that pykerberos is just a fork of kerberos. Or am I missing something?

Lukasa:
@toabctl The switch to PyKerberos was required because of limited support for Python 3 in the Kerberos library for a long time.

However, we can safely switch back.

toabctl:
@Lukasa So py3 support is now good in kerberos? Should I prepare a PR ?

Lukasa:
I believe it is, yes.

rbcarson:
There had been multiple requests and attempts at making the switch to PyKerberos: see #32, #38, #52, #56, #57. One of the claims was that PyKerberos is more actively maintained than python-kerberos, which certainly seemed true at the time.

toabctl:
@rbcarson hm. so are you saying that we should stay with pykerberos? And is it just changing the requirements.txt line or is there more todo for a switch?

rbcarson:
@toabctl I have no preference. Are you able to test the change? We don't have any automated tests against a real server.

In addition to requirements.txt, it's also listed in setup.py's `extras_require`.

behackett:
Apple is [developing the kerberos package again](http://www.calendarserver.org/log/PyKerberos/trunk?rev=15501). They [just did a release on pypi](https://pypi.python.org/pypi/kerberos) in January. It appears to fully support python 3 now.

behackett:
Scratch that. You can't currently switch to Apple's kerberos module. The context object it creates has to be explicitly freed by calling authGSSClientClean when you are done with it. Sadly, that causes issues under python 3.x:

https://www.calendarserver.org/ticket/936

rbcarson:
Note #76 from yesterday, in which Apple's Kerberos 1.2.4 has a bug involving the explicit principal support that was recently added to requests-kerberos (it would have been possible to avoid the issue with some refactoring requests-kerberos in though).

danc86:
Is there any open issue against Apple's Kerberos for the problem observed in #76?

rbcarson:
Not that I know of, though their issue tracker is tricky to dig through. I don't think it's fixed in trunk.

[...]

PR #106, opened 2017-12-07
Change  pykerberos to kerberos

MortalViews:
pykerberos package has gone on life support. changing dependency to original kerberos package. they are both same. 

> NOTE: this fork of ccs-kerberos is currently on life support mode as Apple has resumed work on upstream. Please try to use https://pypi.python.org/pypi/kerberos instead of this fork if possible.

https://github.com/02strich/pykerberos

original kerberos package maintained by apple . 
https://pypi.python.org/pypi/kerberos

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 requests>=1.1.0
 winkerberos >= 0.5.0; sys.platform == 'win32'
-pykerberos >= 1.1.8, < 2.0.0; sys.platform != 'win32'
+kerberos >= 1.1.8, < 2.0.0; sys.platform != 'win32'
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ setup(
     ],
     extras_require={
         ':sys_platform=="win32"': ['winkerberos>=0.5.0'],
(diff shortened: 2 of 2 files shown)


nitzmahone:
A couple of us on the Ansible team have recently taken over maintainership of pykerberos, with the intent to sync it back up with the current Apple kerberos code. That said, we're about to release an update to both pykerberos and requests-kerberos to support Kerberos message encryption. 

We'd prefer to hold off on this change to see how responsive the Apple team is to PRs and releases on kerberos (we'll be submitting our changes in a PR soon). We're not anticipating the further addition of new features, but if there are problems found, we need to be able to patch and release updates quickly.

Once we sync pykerberos back up with the Apple code, keeping the fork around and requests-kerberos using it allows a nice "escape hatch" in case we need to release something faster than the Apple folks want to.

Happy to discuss further, but wanted to make sure you knew the whole story.

MortalViews:
thanks @nitzmahone , my bad i should have looked at the open issues #97 #63 
in [apache airflow](https://github.com/apache/incubator-airflow/blob/master/setup.py#L161) both pykerberos and kerberos is listed as dependences. from what i gather we should be only using pykerberos.

Issue #96, opened 2017-05-16
GSSError: SSPI: InitializeSecurityContext: The specified target is unknown or unreachable


andrewpong:
With `0.11.0` (`requests == 2.14.2`, `winkerberos == 0.6.0`):

I'm receiving the following issue when trying to connect to a REST API on Windows - it works on MacOS, so it seems to be a change in how Kerberos requests are handled on Windows.

Can anyone shed light on why `generate_request_header()` is failing on `authGSSClientStep()`, with `GSSError: SSPI: InitializeSecurityContext: The specified target is unknown or unreachable`?

Is the specified target the host I'm trying to connect to, or the KDC? Is there more detail I can have returned?

```pycon
Unauthorized: Authorization denied - incorrect username or password.

In [3]: webapi = osisoftpy.webapi('https://dev.dstcontrols.com/piwebapi/', authtype=kerberos, verifyssl=False)
[kerberos_.py:188 handle_401() ] D handle_401(): Handling: 401
[kerberos_.py:146 generate_request_header() ] E generate_request_header(): authGSSClientStep() Failed:
Traceback (most recent call last):
  File "D:\venv\lib\site-packages\requests_kerberos\kerberos_.py", line 134, in generate_request_header negotiate_resp_ualue)
GSSError: SSPI: InitializeSecurityContext: The specified target is unknown or unreachable

[kerberos_.py:147 generate_request_header() ] E SSPI: InitializeSecurityContext: The specified target is unknown or unreachable
Traceback (most recent call last):
  File "D:\venv\lib\site-packages\requests_kerberos\kerberos_.py", line 134, in generate_request_header negotiate_resp_ualue)
GSSError: SSPI: InitializeSecurityContext: The specified target is unknown or unreachable

[kerberos_.py:191 handle_401() ] D handle_401(): returning <Response [401]>
[kerberos_.py:284 handle_response() ] D handle_response(): returning <Response [401]>
[kerberos_.py:285 handle_response() ] D handle_response() has seen 0 401 responses
[kerberos_.py:188 handle_401() ] D handle_401(): Handling: 401
[kerberos_.py:146 generate_request_header() ] E generate_request_header(): authGSSClientStep() Failed:
Traceback (most recent call last):
  File "D:\venv\lib\site-packages\requests_kerberos\kerberos_.py", line 134, in generate_request_header negotiate_resp_value)
GSSError: SSPI: InitializeSecurityContext: The specified target is unknown or unreachable

[...]

PR #108, opened 2018-01-02
Resolve URL hostname to FQDN when constructing the SPN

batterseapower:
This should resolve at least some instances of #96 and make having to supply hostname_override a much rarer proposition.

### DIFF
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -3,6 +3,7 @@ try:
 except ImportError:
     import winkerberos as kerberos
 import logging
+import socket
 import re
 import sys
 import warnings
@@ -206,7 +207,16 @@ class HTTPKerberosAuth(AuthBase):
             # allows use of an arbitrary hostname for the kerberos exchange
             # (eg, in cases of aliased hosts, internal vs external, CNAMEs
             # w/ name-based HTTP hosting)
-            kerb_host = self.hostname_override if self.hostname_override is not None else host
+            if self.hostname_override is not None:
+                kerb_host = self.hostname_override
+            else:
+                # We should use the FQDN. This is necessary to support use
+                # cases where multiple services are hosted on the same machine
+                # under multiple CNAMEs. It won't do any good to use a SPN that
+                # mentions one of the CNAMES, we need to ask for the real machine.
+                #
+                # https://chromium.googlesource.com/chromium/src/+/lkgr/net/http/http_auth_handler_negotiate.cc#150
+                kerb_host = socket.getfqdn(host)
             kerb_spn = "{0}@{1}".format(self.service, kerb_host)
 
             result, self.context[host] = kerberos.authGSSClientInit(kerb_spn,
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -162,6 +162,35 @@ def _get_channel_bindings_application_data(response):
 
     return application_data
 
+
+# Return the hostname as it will be canonicalized by
+# krb5_sname_to_principal.  We can't simply use socket.getfqdn()
+# because it explicitly prefers results containing periods and
+# krb5_sname_to_principal doesn't care.
+#
+# This is necessary to support use cases where multiple services
+# are hosted on the same machine under multiple CNAMEs. It won't do
+# any good to use a SPN that mentions one of the CNAMES, we need to ask for the real machine.
+#
+# Maybe we should use socket.getfqdn here, but that doesn't use AI_CANONNAME
+# but rather chooses the first entry in gethostbyaddr that contains a dot, which seems
+# a trifle strange: https://github.com/python/cpython/blob/master/Lib/socket.py#L681
+#
+# https://github.com/krb5/krb5/blob/d406afa363554097ac48646a29249c04f498c88e/src/util/k5test.py#L505-L520
+# https://chromium.googlesource.com/chromium/src/+/lkgr/net/http/http_auth_handler_negotiate.cc#150
+def _get_default_kerb_host(hostname):
+    try:
+        ais = socket.getaddrinfo(hostname, None, 0, 0, 0, socket.AI_CANONNAME)
+    except socket.gaierror:
+        ais = []
+    
+    if not ais:
+        return hostname.lower()
+
+    _family, _socktype, _proto, canonname, _sockaddr = ais[0]
+    return canonname.lower()
+
+
 class HTTPKerberosAuth(AuthBase):
     """Attaches HTTP GSSAPI/Kerberos Authentication to the given Request
     object."""
@@ -210,13 +239,7 @@ class HTTPKerberosAuth(AuthBase):
             if self.hostname_override is not None:
                 kerb_host = self.hostname_override
             else:
-                # We should use the FQDN. This is necessary to support use
-                # cases where multiple services are hosted on the same machine
-                # under multiple CNAMEs. It won't do any good to use a SPN that
-                # mentions one of the CNAMES, we need to ask for the real machine.
-                #
-                # https://chromium.googlesource.com/chromium/src/+/lkgr/net/http/http_auth_handler_negotiate.cc#150
-                kerb_host = socket.getfqdn(host)
+                kerb_host = _get_default_kerb_host(host)
             kerb_spn = "{0}@{1}".format(self.service, kerb_host)
 
             result, self.context[host] = kerberos.authGSSClientInit(kerb_spn,
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -198,7 +198,8 @@ class HTTPKerberosAuth(AuthBase):
             self, mutual_authentication=REQUIRED,
             service="HTTP", delegate=False, force_preemptive=False,
             principal=None, hostname_override=None,
-            sanitize_mutual_error_response=True, send_cbt=True):
+            sanitize_mutual_error_response=True, send_cbt=True,
+            canonicalize_hostname=False):
         self.context = {}
         self.mutual_authentication = mutual_authentication
         self.delegate = delegate
@@ -207,6 +208,7 @@ class HTTPKerberosAuth(AuthBase):
         self.force_preemptive = force_preemptive
         self.principal = principal
         self.hostname_override = hostname_override
+        self.canonicalize_hostname = canonicalize_hostname
         self.sanitize_mutual_error_response = sanitize_mutual_error_response
         self.auth_done = False
         self.winrm_encryption_available = hasattr(kerberos, 'authGSSWinRMEncryptMessage')
@@ -238,8 +240,10 @@ class HTTPKerberosAuth(AuthBase):
             # w/ name-based HTTP hosting)
             if self.hostname_override is not None:
                 kerb_host = self.hostname_override
-            else:
+            elif self.canonicalize_hostname:
                 kerb_host = _get_default_kerb_host(host)
+            else:
+                kerb_host = host
             kerb_spn = "{0}@{1}".format(self.service, kerb_host)
 
             result, self.context[host] = kerberos.authGSSClientInit(kerb_spn,
--- a/README.rst
+++ b/README.rst
@@ -119,6 +119,24 @@ setting the ``hostname_override`` arg:
     >>> r = requests.get("https://externalhost.example.org/", auth=kerberos_auth)
     ...
 
+If ``externalhost.example.org`` is a e.g. CNAME that can be resolved to
+``internalhost.local`` by via DNS, then you can have ``requests-kerberos``
+use the resolved version of the hostname for Kerberos by specifying the
+``canonicalize_hostname`` flag:
+
+.. code-block:: python
+
+    >>> import requests
+    >>> from requests_kerberos import HTTPKerberosAuth
+    >>> kerberos_auth = HTTPKerberosAuth(canonicalize_hostname=True)
+    >>> r = requests.get("https://externalhost.example.org/", auth=kerberos_auth)
+    ...
+
+Canonicalizing hostnames in this way is the default behaviour for almost all browsers.
+We don't match this behaviour because it can weaken security in the case where
+mutual authentication is being used over unauthenticated HTTP to validate the identity
+of the server.
+
 Explicit Principal
 ------------------
 
--- a/README.rst
+++ b/README.rst
@@ -119,8 +119,8 @@ setting the ``hostname_override`` arg:
     >>> r = requests.get("https://externalhost.example.org/", auth=kerberos_auth)
     ...
 
-If ``externalhost.example.org`` is a e.g. CNAME that can be resolved to
-``internalhost.local`` by via DNS, then you can have ``requests-kerberos``
+If ``externalhost.example.org`` is a CNAME that can be resolved to
+``internalhost.local`` via DNS, then you can have ``requests-kerberos``
 use the resolved version of the hostname for Kerberos by specifying the
 ``canonicalize_hostname`` flag:
 
--- a/tests/test_requests_kerberos.py
+++ b/tests/test_requests_kerberos.py
@@ -668,7 +668,7 @@ class KerberosTestCase(unittest.TestCase):
                 response.url = "http://www.example.org/"
                 response.headers = {'www-authenticate': 'negotiate token'}
                 host = urlparse(response.url).hostname
-                auth = requests_kerberos.HTTPKerberosAuth()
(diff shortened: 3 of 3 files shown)


mkomitee:
This should not be merged because it requires that DNS be a trusted source of information. Whether it is or not depends on the administrators of the hosts on which it runs, and if they decide that dns based canonicalization should be used when determining what name to authenticate to, they should be able to configure their kerberos libraries to do so.

In mit kerberos, I believe this policy can be set using the dns_canonicalize_hostname and rdns options.

Figuring out what thing a client intends to authenticate to is a security decision, and choosing for them within the library isn't something we should be doing.

mkomitee:
I just took a look at the referenced issue, and I'd be amendable to a patch like this if it only targeted windows, where there are no such toggles. I'd point out that the referenced code in mit's krb5/sn2princ.c includes a check to the aforementioned `dns_canonicalize_hostname` flag. heimdal has a similar flag which is also used in its `lib/krb5/expand_hostname.c`

https://github.com/krb5/krb5/blob/9c18f6c93b602ec69ca2640e27ab2ccbb7f45a3c/src/lib/krb5/os/sn2princ.c#L68-L87

https://github.com/heimdal/heimdal/blob/37979747bc03ee259ed14c7ac47883074a49a67a/lib/krb5/expand_hostname.c#L71

batterseapower:
If this is a security issue, it doesn't seem like it can be a major one given that (IME, Windows only) all the major browsers do it?

mkomitee:
We can make it a non-default option to canonicalize names, but i know of
companies that are intentional in their kerberos configurations in this
regard. If dnssec was ubiquitous, things might be different.

For the same reason requests validates TLS certificates by default and we
require mutual authentication by default, we shouldn't canonicalize by
default.

mkomitee:
I'll also mention that the comment:

```
# This is necessary to support use cases where multiple services
# are hosted on the same machine under multiple CNAMEs. It won't do
# any good to use a SPN that mentions one of the CNAMES, we need to ask for the real machine.
```

is ... indicative of where the disagreement comes from.

When I am sending a request & want to authenticate to a service, I don't care what the "real machine" is. I care that the service which responds is the service I requested, and so if the service doesn't have access to the key which it could use to prove that it is the service that  I expect it to be, I want a mutual authentication error.

With this patch in place, if I want to send a request to trusted.com, and someone wants to pretend to be trusted.com to me, all they'd need to do is find a way to get me to receive a CNAME for trusted.com to redirect me to malicious.com instead. I could then successfully authenticate against malicious.com. All malicious.com would have to do is be able to prove that it was malicious.com for me to trust them. I'd be none-the-wiser.

Also, with apache & nginx (and many other web servers which support kerberos), if they're configured with a single keytab file that has access to keys for all of their named virtual hosts, they'll pick the right one based on the Host header on the request, so it does do good to use the SPN that mentions the CNAME, not the real machine.

batterseapower:
Isn't the scenario you describe really only relevant for Requests if your server is not using HTTPS? If DNS is untrustworthy, then once you've received your Kerberos authentication token your HTTP request may end up in the hands of a malicious actor regardless of how careful we were at the Kerberos stage.

Similarly, if the server **does** use HTTPS then I don't see why it's problematic to send a token generated for malicious.com to trusted.com. Sure, authentication will fail, but that's expected.

So it's not clear to me that being extra strict at the Kerberos stage really buys us anything in this context? Perhaps I misunderstand.

I have no objections to putting this under flag control, but I strongly feel that the default behaviour should be to canonicalise. Reasons:

 1. The default value in krb5 itself is true: https://github.com/krb5/krb5/blob/a4ddc6cf576b4155e6b994307902567f26f752b2/doc/admin/conf_files/krb5_conf.rst
 2. Requests is "HTTP for humans". I think humans will expect Requests to behave like the browsers do by default, which means we should canonicalise. (This is also why we should validate TLS certs by default.)

mkomitee:
Several RFCs explicitly point out the security issue created by canonicalizing hostnames in GSSAPI using DNS.

 For example, RFC4462 (for SSH GSS-API methods) says this:

>  Because the GSS-API mechanism uses the targ_name to authenticate the
>    server's identity, it is important that it be determined in a secure
>    fashion.  One common way to do this is to construct the targ_name
>    from the hostname as typed by the user; unfortunately, because some
>    GSS-API mechanisms do not canonicalize hostnames, it is likely that
>    this technique will fail if the user has not typed a fully-qualified,
>    canonical hostname.  Thus, implementers may wish to use other
>    methods, but should take care to ensure they are secure.  For
>    example, one should not rely on an unprotected DNS record to map a
>    host alias to the primary name of a server, or an IP address to a
>    hostname, since an attacker can modify the mapping and impersonate
>    the server.
> 
>    Implementations of mechanisms conforming to this document MUST NOT
>    use the results of insecure DNS queries to construct the targ_name.
>    Clients MAY make use of a mapping provided by local configuration or
>    use other secure means to determine the targ_name to be used.  If a
>    client system is unable to securely determine which targ_name to use,
>    then it SHOULD NOT use this mechanism.

RFC4752 (for the Kerberos V5 (GSSAPI) SASL mechanism) has this to say:

> Security issues are discussed throughout this memo.
> 
>    When constructing the input_name_string, the client SHOULD NOT
>    canonicalize the server's fully qualified domain name using an
>    insecure or untrusted directory service.

It's not explicitly forbidden with GSSAPI itself, but since tools are in place for a properly administered kerberos site to decide to canonicalize or not based on a site wide policy, we should not make a decision which counter-acts that global policy by default.

batterseapower:
Thanks for the pointers to the RFCs, but I still don't understand the threat. Let's assume that DNS is compromised -- what exactly goes wrong if we canonicalize to malicious.com before requesting the security context?

We end up sending the token for malicious.com to either malicious.com or good.com (depending on what the person who controls DNS desires).
  - If the token is sent to good.com then authentication fails, so there is no harm done (it's not a new DoS attack because the DNS controlled could always blackhole your request to some bad IP anyway).
  - If the token is sent to malicious.com then the malicious guy gets your token for malicious.com, but this can't be a problem. (Reasoning: if you control DNS then even if krb does no canonicalisation, you could always have the client generate a token for good.com and have it sent to malicious.com. If **that** isn't a problem, then malicious.com getting a token for their own malicious service cannot be a problem.)

So, again, I don't really see the threat? Like I say - I might just be missing something.

I appreciate that the RFCs use the "MUST/SHOULD NOT" language, but standards are not always a good guide about how to build a functioning member of the web ecosystem, and this case accepted practice for Kerberos auth in the web context clearly differs from what these documents say.

Finally, whether we choose to canonicalize or not, we are making **some** policy choice which may conflict with the accepted policy at the use site. I believe the onus of "hardening" Kerberos should be placed on the 0.1% of users who understand the security issues involved and can make an informed choice to turn canonicalization off in this lib. I do not think we should impose the cost of explicitly configuring requests to behave like a browser on the 99.9% of users who reasonably expect that behaviour to be the default.

mkomitee:
It’s not about trusting malicious.com with your token. The only thing the
token is useful for us pricing your identity.

It's about whether the user can trust the contents of the response which
the user would believe came from good.com since mutual authentication would
have succeeded.

batterseapower:
OK, I think I see - thanks for explaining. So there is a narrow range of situations where merging this PR would reduce security. Specifically that would require all of the following to be true:

 - Malicious actor controls DNS
 - HTTPS is not being used (if HTTPS were used, that would establish the identity of the server independently of the mutual auth mechanism)
 - Malicious actor *not* able to modify the HTTP response (I think that with enough control of the network, an attacker could replace the whole body of the response after the WWW-Authenticate header with their own bogus response, so validity of WWW-Authenticate tells you nothing -- you really need the HTTPS message integrity check)

This seems like a pretty narrow range of scenarios? But, I guess maybe someone cares about them, and someone could argue that the potential security loss in those situations outweighs the inconvenience of having to reenable canonicalisation in the relevant client code, to restore browser compatibility.

In any event it seems we can both agree that we can do this canonicalisation by default if mutual_authentication is DISABLED.

mkomitee:
You're right about the fact that using HTTP without HTTPS is a bad idea, but people do that and may rely on Kerberos w/ mutual authentication to protect them as much as it's able.

Another problem is that enabling canonicalization by default could break things for people which are currently working. If they've configured their kerberos environment to not canonicalize names using DNS they're relying on their servers to have access to keys for the CNAME rather than the target of the CNAME.

Also, canonicalization can mean different things to different people. Some environments canonicalize by resolving CNAMEs (dns_canonicalize_hostname) , while others canonicalize by resolving CNAMEs and PTRs (rdns).

I don't think it's a good idea to change the default in a way that takes away an environment admins ability to configure their environment, even in the case of mutual authentication being disabled. 

... as a compromise, what about the possibility of making the defaults configurable within the library? Meaning, we make 2 module level constants, DEFAULT_DNS_CANONICALIZE_HOSTNAME and DEFAULT_RDNS, set them both to False by default, and add two constructor arguments on HTTPKerberosAuth named dns_canonicalize_hostname and rdns, and have them default to whatever the module level constant values are.

With them both off (the default), it will defer to whatever the sites sysadmins configured their kerberos libraries to do, but you can change the default once in the module so you don't need to do it independently for every instance created.

batterseapower:
I've put this under flag control. I didn't add a module level flag because I don't think that will significantly reduce the burden of setting the flag -- in all my apps I just have one bit of code that creates the HTTPKerberosAuth object, then reuse this in all context where I use requests.

mkomitee:
Sorry I haven't gotten back to you on this. I'm happy with this change, but
I think some of the wording in the doc needs a going over.

It's probably enough to drop "e.g." and "by"

requests_kerberos/kerberos_.py
```diff
@@ -161,14 +162,44 @@ def _get_channel_bindings_application_data(response):
...
+def _get_default_kerb_host(hostname):
+    try:
+        ais = socket.getaddrinfo(hostname, None, 0, 0, 0, socket.AI_CANONNAME)
+    except socket.gaierror:
+        ais = []
+    
+    if not ais:
+        return hostname.lower()
+
+    _family, _socktype, _proto, canonname, _sockaddr = ais[0]
```
mkomitee: Based on how we intend to use this, I'm wondering if we should filter out anything that has a family which isn't AF_INET or AF_INET6, as well as anything which has a socktype which isn't SOCK_STREAM, as well as anything which doesn't have a photo of IPPROTO_TCP.

Do you think that's paranoid? 

I'm not certain whether or not cannoname is expected to be the same for all results in the ais list. It is for everything I tried, but that's different from verification of the spec.

requests_kerberos/kerberos_.py
```diff
@@ -161,14 +162,44 @@ def _get_channel_bindings_application_data(response):
...
 class HTTPKerberosAuth(AuthBase):
     """Attaches HTTP GSSAPI/Kerberos Authentication to the given Request
     object."""
     def __init__(
             self, mutual_authentication=REQUIRED,
             service="HTTP", delegate=False, force_preemptive=False,
             principal=None, hostname_override=None,
-            sanitize_mutual_error_response=True, send_cbt=True):
+            sanitize_mutual_error_response=True, send_cbt=True,
+            canonicalize_hostname=False):
```
mkomitee: I can foresee someone asking us to also reverse canonicalize this in the future. I'm not saying you should implement that, but naming the argument canonicalize_hostname and having it a boolean can make the API messy in the future if we do add that.

Maybe we should make it an enumerable of sorts, like mutual_authentication. The default can switch from False to None, and we can declare a module level constant like REQUIRED/OPTIONAL/DISABLED, but named FORWARD. That leaves us the option to later make another option named REVERSE.

requests_kerberos/kerberos_.py
batterseapower: I'm not sure that we should do extra filtering. By doing so I think we're just as likely to break some obscure deployment as we are to fix any problems. What if someone is still using IPX networking, or running requests-kerberos in the year 2045 over IPv7?

I couldn't find any definitive answer about this, but it does seem to me that canonname should be the same across all entries. It's a property of the host, not the particular socket/address that is used to communicate with it.

stphnlyd:
Hi @mkomitee, @batterseapower 

At my workspace we got the similar issue as #96 on Windows with requests-kerberos + winkerberos. And we would like to see in requests-kerberos at least of some level of support for resolving FQDN. 

Now reading this thread here I wonder what status is of this PR. Are you going to continue working on it? 

And, I am not sure if it's a good idea but if there turns out to be still a lot details to discuss as you might want to cover , I wonder if we can have a alternative solution like this: allow that hostname_override to be a callable (so it can be either a string or a callable), the callable is like f(hostname) -> fqdn, and leave the implementation to users of this library?

batterseapower:
@stphnlyd I'm not pushing this forward any longer because my aim was to make canonicalization the default behaviour so that client code written by my webservice consumers would automatically do the right thing. Since the maintainer doesn't want canonicalization to be on by default, this means whether or not this PR is merged I'll _still_ have to educate users of my webservices about the need for canonicalization, and in this case it's just as much work for me to tell them to stick in a `hostname_override` as it is to use the proposed `canonicalize_hostname`.

I still believe that it should be the default behaviour: the potential security hole outlined above doesn't seem worth worrying about, and it's surprising that the default behaviour of requests-kerberos would differ from that of browsers.

That being said, in my opinion this PR is ready to merge as-is? @mkomitee said he would be happy to merge it after a doc tweak, and I already made the tweak he asked for above.

stphnlyd:
@batterseapower thanks for your work and this reply. I am neutral regarding if it's default or not: yes it's best to be default like browsers, but if the author have much concerns that's also fine to not be default. But I really need a simple mechanism to enable auto FQDN resolving, that will be much better than present string-typed `hostname_override`.

@mkomitee can you see this?

zhouzhen1:
At my workplace we also have this problem. Can someone take a look at this thread please?

@mkomitee , @nitzmahone , @jborean93

jborean93:
Closing as resolving the hostname from the IP is fraught with danger and not something I think requests-kerberos should really be doing. There are config options for `/etc/krb5.conf` that you can tweak to enable this behaviour on the GSSAPI land but if the caller wishes to resolve the hostname they should be using the `hostname_override` option and requests-kerberos just sticks with the current behaviour.

batterseapower:
@jborean93 thanks for looking at the request again. Even if you believe that doing this is "fraught with danger", can we agree that having a default-off option to do this isn't so bad so long as the docs call out the potential problems? `HTTPKerberosAuth` already allows you to configure it in some ways that you would presumably consider dangerous (e.g. `mutual_authentication=OPTIONAL`).

I think if you force people to use `hostname_override`, almost all users are just going to copy what e.g. Chrome does (https://github.com/chromium/chromium/blob/72ceeed2ebcd505b8d8205ed7354e862b871995e/net/http/http_auth_handler_negotiate.cc#L339) and implement exactly the proposed DNS lookup, so you aren't really making these systems any more secure by forcing this lookup to be made from user code rather than from your library.

Thanks for mentioning that `krb5.conf` could in some cases be changed to do this resolution for you - I didn't know this. This is interesting, but it doesn't help those who want their applications to work on Windows, or on systems where the users can't control the krb config.

jborean93:
Admittedly I misread the original problem and thought this was about resolving an IP -> DNS name but after looking deeper I still don't believe this is something that should be implemented here.

My main opposition is really this library doing any DNS resolutions itself and the points mentioned in https://github.com/requests/requests-kerberos/pull/108#issuecomment-374804424. In reality the proper method IMO is to have the services behind the CNAME share the same keytab which is registered with the SPN `HTTP/my-cname`. While Windows does not support keytabs through SSPI it is still possible by having those services run using a shared domain account with the CNAME registered on that account's SPNs. The typical way this is done is through a group managed service account.

My 2 other concerns with the changes (even when behind an opt in flag are):

* I don't know how portable `socket.getaddrinfo` is across the various platforms
  * My knowledge on this is pretty lax but I've encountered enough `socket` differences across the platforms, especially Windows vs Linux
* If you are trying to resolve a `CNAME` what's stopping this `getaddrinfo` call returning 1 hostname but the underlying requests call on it's own socket resolving another hostname
  * This brings us back to the beginning where the ticket is encrypted for 1 host but the actual host requests has connected 2 is another
  * Maybe caching helps in this regard but I would be very wary to definitively say it's not a problem

In the end there is already a mechanism to override the hostname portion used in the SPN and the caller is free to use this function to "resolve" the hostname on their code and pass that into `requests-kerberos`. It might be extra code on their side but they would know more details about the host they are running on and their environment and there might be a better method for them to use. This coupled with the fact that I personally believe this sort of problem should be solved by sharing the keytab/service principal on the alias targets instead of through client side workarounds means that it's not something that should be implemented here.

I understand this is not what some people like to hear and I do apologise for having the PR lapse for so long but I hope my reasoning makes some reason.

PR #115, opened 2018-04-10
allow to specify a mech_oid to authGSSClientInit

pu239ppy:
Seems that for certain server endpoints we use (Apache Jetty) require that we specify mech OID to `authGSSClientInit`.  This is a simpler follow up to #89 that does not change default behavior if such change is not explicitly asked for.

This is a minimal code change, in essence if mech_oid is not specified in the constructor the same logic will be executed as before 

````python
result, self.context[host] = kerberos.authGSSClientInit(kerb_spn, gssflags=gssflags, principal=self.principal)
````
However if `mech_oid` is passed to the constructor the following is executed 
````python
result, self.context[host] = kerberos.authGSSClientInit(kerb_spn,
gssflags=gssflags, principal=self.principal, mech_oid=mech_oid)
````
It is up to the caller to find the right mech_oid (e.g. `mech_oid=kerberos.GSS_MECH_OID_SPNEGO` the requests-kerberos library does not expose/re-export anything to the client

### DIFF
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -167,7 +167,7 @@ class HTTPKerberosAuth(AuthBase):
     def __init__(
             self, mutual_authentication=REQUIRED,
             service="HTTP", delegate=False, force_preemptive=False,
-            principal=None, hostname_override=None,
+            principal=None, hostname_override=None, mech_oid=None,
             sanitize_mutual_error_response=True, send_cbt=True):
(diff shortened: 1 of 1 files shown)


pu239ppy:
Any chance this may get reviewed/merged soon?

pu239ppy:
Repo maintainers, any way this can get merged soon?  I really dislike using a forked version of this.  As previously stated the PR does not change default behavior.

PR #118, opened 2018-06-13
Support passing in context with delegated credentials

jeffreythewang:
In order to fully support Kerberos delegation, a service should be able to authenticate with Kerberos on behalf of a user. That user's credentials can be passed in to the `delegated` parameter of `kerberos.authGSSClientInit` as documented [here](https://github.com/apple/ccs-pykerberos/blob/master/pysrc/kerberos.py#L159).

I've updated the `README` to include a scenario of how a context with delegated credentials can be initialized. Note that I pass in `"DELEGATE"` as the service principal, which is documented [here](https://github.com/apple/ccs-pykerberos/blob/master/pysrc/kerberos.py#L353).

I realized this requires users of the interface to interact with GSS API more directly, and I'm open to better ways of abstracting this away, if possible.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -147,13 +147,17 @@ Delegation
 
 ``requests_kerberos`` supports credential delegation (``GSS_C_DELEG_FLAG``).
 To enable delegation of credentials to a server that requests delegation, pass
-``delegate=True`` to ``HTTPKerberosAuth``:
+``delegate=True`` and ``delegated_context=<context>`` (where context is the server
+context containing delegated credentials) to ``HTTPKerberosAuth``:
 
 .. code-block:: python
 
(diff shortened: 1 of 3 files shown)


behackett:
Note that this won't work with WinKerberos. It doesn't look like SSPI supports an API to do this, even after WinKerberos gains support for server side GSSAPI (currently available in a pull request).

mkomitee:
Can you add some test coverage which exercises delegation?

Also, can you add something to fail fast when delegation is attempted but not supported by the platform (e.g. winkerberos)?

jborean93:
This PR is no longer works after merging https://github.com/requests/requests-kerberos/pull/163.

PR #121, opened 2018-07-05
Fix invalid escape sequence "\s"

frozencemetery:
Mark the regex as a raw string instead.

Also-authored-by: Stanislav Levin <slev@altlinux.org>

See: https://github.com/pythongssapi/requests-gssapi/pull/10 and https://github.com/pythongssapi/requests-gssapi/issues/9


dsully:
Can this be committed and pushed to pypi? Thanks.

mkomitee:
Thanks for your contribution.

PR #123, opened 2018-08-02
Added missing trove classifier for license

### LINKED ISSUE #122: license missing from pypi
Apparently https://pypi.org/project/requests-kerberos/ is missing to display license metadata.

Ideally requests-kerberos should have the same license as requests as I don't see it used separately
but that's outside the scope of this ticket.

I went into this when I enabled https://pyup.io/account/repos/github/pycontribs/jira/ to check license dependencies and reported that requests-kerberos does not have a known license or OSI approved.

ssbarnea:
Now PYPI should be able to recognise license.

Fixes #122

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -58,4 +58,7 @@ setup(
     },
     test_suite='test_requests_kerberos',
     tests_require=['mock'],
+    classifiers=[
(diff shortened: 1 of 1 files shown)


carsonyl:
Thank you. :)

PR #126, opened 2018-12-21
Update requirements to use kerberos instead of pykerberos

henry0312:
https://github.com/02strich/pykerberos (https://pypi.org/project/pykerberos/) says
"NOTE: this fork of ccs-kerberos is currently on life support mode as Apple has resumed work on upstream.
Please try to use https://pypi.python.org/pypi/kerberos instead of this fork if possible.".
So, this PR will switch pykerberos to https://github.com/apple/ccs-pykerberos (https://pypi.org/project/kerberos/).

PR #127, opened 2018-12-21
Update .travis.yml to drop Python 3.3 and support Python 3.7

henry0312:


mkomitee:
I'm not familiar with travis. What's the reason for switching to xenial?

Is there any reason to stop testing 3.3?

 It would be good to know when/if we want to merge a change that breaks 3.3 support. When that happens we can make the decision to officially drop support.

mkomitee:
Oh, I see it already doesn't work from https://github.com/requests/requests-kerberos/pull/126.

henry0312:
@mkomitee Because tests with Python 3.7 on Travis CI requires xenial.

mkomitee:
Thanks for your contribution.

PR #129, opened 2019-02-13
Try winkerberos first

tcely:
In the previous order, the `ImportError` is masked on Linux and leads to a confusing message about not finding the `winkerberos` which just confuses most Linux users.

This ordering lets the real issue be exposed for Linux users. Windows users are less likely to be missing pieces and there should be fewer users of this software on Windows too.

### DIFF
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -1,7 +1,7 @@
 try:
-    import kerberos
-except ImportError:
     import winkerberos as kerberos
+except ImportError:
+    import kerberos
 import logging
 import re
 import sys
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -1,10 +1,16 @@
-try:
-    import winkerberos as kerberos
-except ImportError:
-    import kerberos
+import sys
+if sys.platform in ["win32", "cygwin"]:
+    try:
+        import kerberos
+    except ImportError:
+        import winkerberos as kerberos
+else:
+    try:
+        import winkerberos as kerberos
+    except ImportError:
+        import kerberos
 import logging
 import re
-import sys
 import warnings
 
 from cryptography import x509


mkomitee:
I could see that as a breaking change for anyone on cygwin who has a linux like kerberos environment around.

Instead of trying one and then another which will lead to confusion for *someone*, would it be better to explicitly check sys.platform, and use winkerberos if it's win32?

Does anyone with windows / cygwin know what sys.platform returns?

jon-turney:
> Does anyone with windows / cygwin know what sys.platform returns?

```
$ uname -a
CYGWIN_NT-10.0 tambora 2.11.2(0.329/5/3) 2018-11-08 14:34 x86_64 Cygwin

$ python3 -c 'import sys; print(sys.platform)'
cygwin
```

mkomitee:
so a platform check would work.

behackett:
Does cygwin support krb5, and can you build ccs-pykerberos there? MSYS2 build support was recently added to WinKerberos in https://github.com/mongodb-labs/winkerberos/pull/22

mkomitee:
I believe there are mit-krb5 libraries available for cygwin, and I'm assuming that means `kerberos` would be preferred default on the platform.

So I'm thinking we can do a platform check, use that to determine the correct order to try. We'll still try both, but if the sys.platform indicates it's windows (non cygwin) we try kerberos then winkerberos so the error message reported is for winkerberos, and on all other platforms we try winkerberos then kerberos so the error message reported is for kerberos.

So in either case, if they have either, it'll still work, but the error message they receive if they have neither makes the most sense for the given platform.

mtdeguzis:
I agree, this led me down a confusing path of discovery when I was switching over a in-house utility to Python 3-only. A platform check I thnk is a good option and/or a better message to the user.

Only when I saw the import logic, did I go "oh, duh, build kerberos from PyPi as an RPM for RHEL first".

```
  1 try:
  2     import kerberos
  3 except ImportError:
  4     import winkerberos as kerberos

    extras_require={
        ':sys_platform=="win32"': ['winkerberos>=0.5.0'],
        ':sys_platform!="win32"': ['pykerberos>=1.1.8,<2.0.0'],
    },
```

```
[mtdeguzis@host: ~]$ python3 -c 'import kerberos;print(kerberos);'                                                                   
<module 'kerberos' from '/usr/lib64/python3.6/site-packages/kerberos.cpython-36m-x86_64-linux-gnu.so'>
[mtdeguzis@host: ~]$ python3 -c 'import requests_kerberos;print(requests_kerberos);'
<module 'requests_kerberos' from '/usr/lib/python3.6/site-packages/requests_kerberos/__init__.py'>
```

mtdeguzis:
Does this patch need refreshed? I don't mind, but I leave it up to the author and the project. This would definitely help out CentOS / RHEL folks doing rebuilds for Python 3.

jborean93:
This is no longer relevant with the change https://github.com/requests/requests-kerberos/pull/163. The `winkerberos` library is no longer being used and is replaced by `pyspnego` which is a cross platform implementation of GSSAPI and SSPI.

PR #131, opened 2019-05-24
Update readme

sblcook:
I added a "setup" section to the README. When I first used the package, I had several questions that were not answered in the README, so I figured that I could update to help anyone with future questions.


mkomitee:
Thanks for your contribution.

PR #138, opened 2019-10-22
Do not log Authorization header values.

### LINKED ISSUE #137: Do not log Authorization headers
Authorization headers may contain sensitive information that should not be logged (even at DEBUG level). requests-kerberos is currently logging Authorization headers in the following places:

- https://github.com/requests/requests-kerberos/blob/d9c78fc3040481bf944306193ee5564ae6c4de31/requests_kerberos/kerberos_.py#L439
- https://github.com/requests/requests-kerberos/blob/d9c78fc3040481bf944306193ee5564ae6c4de31/requests_kerberos/kerberos_.py#L266-L267

Would you accept a patch that removes the contents of the header from these log messages?

Thanks for your consideration.

twosigmajab:
Fixes #137.

### DIFF
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -263,8 +263,7 @@ class HTTPKerberosAuth(AuthBase):
             # GSS Failure, return existing response
             return response
 
-        log.debug("authenticate_user(): Authorization header: {0}".format(
-            auth_header))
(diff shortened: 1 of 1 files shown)


jborean93:
The sensitive values in these headers are encrypted by the session that created them and can be helpful for debugging purposes so I don't think omitting them is really too helpful. In reality if you don't want these in your log file disable debug logging for this particular library.

PR #147, opened 2020-02-26
3.8 is out. Remove EOL versions and bump the rest

ecederstrand:


ecederstrand:
Ping?

ecederstrand:
@jborean93 @nitzmahone or @mkomitee Ping?

jborean93:
I can’t speak for the rest of the maintainers but this package is effectively just on life support from my perspective. Myself and @nitzmahone are really only maintainers for pywinrm/ansible Kerberos support and we have no real desire to make introduce new changes unless they fix bugs we are encountering.

This library as it is today is stable enough to let it keep on running as things aren’t broken yet. If you are wanting something that is new and add more features to I highly recommend using requests-gssapi. This uses a far better gssapi library than pykerberos while offering the same API as requests-kerberos.

FHTMitchell:
> I can’t speak for the rest of the maintainers but this package is effectively just on life support from my perspective. Myself and @nitzmahone are really only maintainers for pywinrm/ansible Kerberos support and we have no real desire to make introduce new changes unless they fix bugs we are encountering.
> 
> This library as it is today is stable enough to let it keep on running as things aren’t broken yet. If you are wanting something that is new and add more features to I highly recommend using requests-gssapi. This uses a far better gssapi library than pykerberos while offering the same API as requests-kerberos.

should this perhaps be put in the readme?

Issue #148, opened 2020-03-09
Proxy Authentication


christophvw:
Could it be that kerberos proxy authentication is not supported yet?

407 - Proxy-Authenticate', 'Proxy-Authorization

christophvw:
patch for proxy-support

TODO:
-host should be set to the proxy server (currently you have to use hostname_override)
-fix urlib3 HTTPS proxy support


`--- C:/Users/wittich/Desktop/kerberos_.py	Mo  9. Mrz 14:04:14 2020
+++ C:/Users/wittich/Desktop/kerberos_proxy.py	Mo  9. Mrz 14:04:52 2020
@@ -88 +88,4 @@ def _negotiate_value(response):
-    authreq = response.headers.get('www-authenticate', None)
+    if response.status_code == 401:
+        authreq = response.headers.get('www-authenticate', None)
+    else:
+        authreq = response.headers.get('proxy-authenticate', None)
@@ -266,3 +269,8 @@ class HTTPKerberosAuth(AuthBase):
-        log.debug("authenticate_user(): Authorization header: {0}".format(
-            auth_header))
-        response.request.headers['Authorization'] = auth_header
+        if response.status_code == 401:
+            log.debug("authenticate_user(): Authorization header: {0}".format(
+                auth_header))
+            response.request.headers['Authorization'] = auth_header
+        elif response.status_code == 407:
+            log.debug("authenticate_user(): Proxy-Authorization header: {0}".format(
+                auth_header))
+            response.request.headers['Proxy-Authorization'] = auth_header
@@ -293,0 +302,13 @@ class HTTPKerberosAuth(AuthBase):
+    def handle_407(self, response, **kwargs):
+        """Handles 407's, attempts to use gssapi/kerberos authentication"""
+
+        log.debug("handle_407(): Handling: 407")
+        if _negotiate_value(response) is not None:
+            _r = self.authenticate_user(response, **kwargs)
+            log.debug("handle_407(): returning {0}".format(_r))
+            return _r
+        else:
+            log.debug("handle_407(): Kerberos is not supported")
+            log.debug("handle_407(): returning {0}".format(response))
+            return response
+
@@ -376,0 +398 @@ class HTTPKerberosAuth(AuthBase):
+        num_407s = kwargs.pop('num_407s', 0)
@@ -409,0 +432,13 @@ class HTTPKerberosAuth(AuthBase):
+        elif response.status_code == 407 and num_407s < 2:
[...]

PR #149, opened 2020-04-07
Add Kerberos proxy authentication handling

enzolis:
This is a PR based on the discussion here:
https://github.com/requests/requests-kerberos/issues/148

Basically, it adds handling of status code 407 and the generation of an authentication header to communicate via an authenticating proxy (in my case a company proxy).


jborean93:
The changes in https://github.com/requests/requests-kerberos/pull/163 has resulted in some merge conflicts in the test file. If you are still interested in getting this in I can have a look at the changes once it has been fixed otherwise I'll close this PR.

jborean93:
Thanks for rebasing the code. I've just had a play with the changes here and I was able to authenticate with a Kerberos backed proxy when connecting to a HTTP endpoint but it seems like this doesn't work for HTTPS endpoint (the final target is HTTPS not the proxy). Requests fails with

> requests.exceptions.ProxyError: HTTPSConnectionPool(host='hostname', port=1234): Max retries exceeded with url: / (Caused by ProxyError('Cannot connect to proxy.', OSError('Tunnel connection failed: 407 Proxy Authentication Required')))

It does seem like this might be a bug in requests where it is not calling the auth handler for failed `CONNECT` requests which is how proxies work when the final target is a HTTPS endpoint. Is this something you can replicate on your end, just trying to rule out whether this is just maybe an environment problem on my side or something that should be called out explicitly.

enzolis:
Hi Jordan,

this is unfortunately correct. I think these contain the relevant information:
https://github.com/psf/requests/issues/1582
https://github.com/urllib3/urllib3/issues/242

BR

jborean93:
Thanks for clarifying, I think for the initial work we can live with that support and add `HTTPS` support in the future if requests eventually fixes that problem.

PR #151, opened 2020-05-29
allow user to pass in mech_oid

DSchmidlin:
Allow the user to pass in an OID to avoid the following error:

SPNEGO negotiation token is not a NegTokenInit: OID 1.2.840.113554.1.2.2 does not match SPNEGO OID 1.3.6.1.5.5.2

passing in kerberos.GSS_MECH_OID_SPNEGO resolves this error.


jborean93:
The changes in https://github.com/requests/requests-kerberos/pull/163 conflict with the changes here due to the change in underlying libraries. Ultimately the `pyspnego` library will be calling GSSAPI with the mech `1.2.840.113554.1.2.2` (Kerberos). This library is designed for Kerberos auth and I'm trying to keep it solely for that protocol rather than trying to add in SPNEGO support.

PR #163, opened 2021-10-28
Change underlying lib to pyspnego

jborean93:
This changes the underlying Kerberos library from `pykerberos` and `winkerberos` to `pyspnego`. The `pyspnego` library handles the platforms differences between Windows and Linux meaning there is now only 1 dependency.

Some other changes being made is designed to hide the underlying library used a bit more to avoid external libraries using it for their own purposes. This will ensure the kerb library can easily be changed in the future if need be.

The tests have been turned into `pytest` tests as a few of the mocking functions needed to be changed due to the new API of `pyspnego` compared to the existing ones.


jborean93:
Have tested the changes from both a Linux, macOS, and Windows host against a Kerberos auth'd HTTP backend without any issues. Will release an rc of `0.13.0` just to be sure but things look like they work.

Issue #168, opened 2021-11-29
"BytesWarning: str() on a bytes instance" in authenticate_server()


etfdev:
I run all my code using `python -bb` (see <https://docs.python.org/3/using/cmdline.html#cmdoption-b>) in order to detect str vs. bytes mixups. Unfortunately this leads to an exception in `requests_kerberos`:

```
$ python -bb -m IPython
In [1]: import requests
   ...: import requests_kerberos
   ...: s = requests.Session()
   ...: s.auth = requests_kerberos.HTTPKerberosAuth(mutual_authentication=requests_kerberos.OPTIONAL, force_preemptive=True)
   ...: s.verify = False
   ...: s.get("<internal server (censored)>")

---------------------------------------------------------------------------
BytesWarning                              Traceback (most recent call last)
<ipython-input-1-ea4faad4cf5c> in <module>
      4 s.auth = requests_kerberos.HTTPKerberosAuth(mutual_authentication=requests_kerberos.OPTIONAL, force_preemptive=True)
      5 s.verify = False
----> 6 s.get("<internal server (censored)>")

/custom-python/lib/python3.7/site-packages/requests/sessions.py in get(self, url, **kwargs)
    541
    542         kwargs.setdefault('allow_redirects', True)
--> 543         return self.request('GET', url, **kwargs)
    544
    545     def options(self, url, **kwargs):

/custom-python/lib/python3.7/site-packages/requests/sessions.py in request(self, method, url, params, data, headers, cookies, files, auth, timeout, allow_redirects, proxies, hooks, stream, verify, cert, json)
    528         }
    529         send_kwargs.update(settings)
--> 530         resp = self.send(prep, **send_kwargs)
    531
    532         return resp

/custom-python/lib/python3.7/site-packages/requests/sessions.py in send(self, request, **kwargs)
    648
    649         # Response manipulation hooks
--> 650         r = dispatch_hook('response', hooks, r, **kwargs)
    651
    652         # Persist cookies

/custom-python/lib/python3.7/site-packages/requests/hooks.py in dispatch_hook(key, hooks, hook_data, **kwargs)
     29             hooks = [hooks]
     30         for hook in hooks:
---> 31             _hook_data = hook(hook_data, **kwargs)
     32             if _hook_data is not None:
     33                 hook_data = _hook_data

/custom-python2/lib/python3.7/site-packages/requests_kerberos/kerberos_.py in handle_response(self, response, **kwargs)
    374             return response
    375         else:
[...]

PR #169, opened 2021-11-29
Fix stray byte conversion and added changelog

jborean93:
The raw Kerberos token from the server was being used in the debug logs which can be problematic when dealing with invalid codepoints. This PR makes sure we log the base64 value which will always contain just ASCII chars.

Also adds the changelog entry for https://github.com/requests/requests-kerberos/pull/149.

Fixes https://github.com/requests/requests-kerberos/issues/168

### DIFF
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -348,13 +348,16 @@ class HTTPKerberosAuth(AuthBase):
         Returns True on success, False on failure.
         """
 
+        response_token = _negotiate_value(response)
         log.debug("authenticate_server(): Authenticate header: {0}".format(
-            _negotiate_value(response)))
+            base64.b64encode(response_token).decode()
+            if response_token
+            else ""))
 
         host = urlparse(response.url).hostname
 
         try:
-            self._context[host].step(in_token=_negotiate_value(response))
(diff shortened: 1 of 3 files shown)


requests_kerberos/kerberos_.py
```diff
@@ -348,13 +348,16 @@ def authenticate_server(self, response):
         Returns True on success, False on failure.
         """
 
+        response_token = _negotiate_value(response)
         log.debug("authenticate_server(): Authenticate header: {0}".format(
-            _negotiate_value(response)))
+            base64.b64encode(response_token).decode()
+            if response_token
+            else ""))
```
etfdev, line 355: Note that `pylint` and other tools will complain about non-lazy logger usage here.

You might want to do

```
log.debug("authenticate_server(): Authenticate header: %s", base64.b64encode(response_token).decode() if response_token else "")
```

With this approach if the debug level isn't high enough your code will avoid one string formatting operation.

To make it even lazier, you can do

```
if log.isEnabledFor(logging.DEBUG):
    log.debug(...)
```

jborean93: Happy to accept a PR if you wish to make the change

PR #182, opened 2023-07-27
fix #181 add password param to constructor

### LINKED ISSUE #181: Defining password in principal does not create a spnego client with password

razzzp:
Hi,

This pull request is for parsing the principal if it is given in the form of user@REALM:password. It will split the username and password and pass them both to spnego.client().

### DIFF
--- a/requests_kerberos/kerberos_.py
+++ b/requests_kerberos/kerberos_.py
@@ -210,26 +210,16 @@ class HTTPKerberosAuth(AuthBase):
             # w/ name-based HTTP hosting)
             kerb_host = self.hostname_override if self.hostname_override is not None else host
 
-            # to prevent failing compare headers test, don't pass password if None
-            if self.password:
-                self._context[host] = ctx = spnego.client(
-                    username=self.principal,
-                    password=self.password,
-                    hostname=kerb_host,
-                    service=self.service,
-                    channel_bindings=self._cbts.get(host, None),
-                    context_req=gssflags,
-                    protocol="kerberos",
-                )
-            else:
-                self._context[host] = ctx = spnego.client(
-                    username=self.principal,
-                    hostname=kerb_host,
-                    service=self.service,
-                    channel_bindings=self._cbts.get(host, None),
-                    context_req=gssflags,
-                    protocol="kerberos",
-                )
+            self._context[host] = ctx = spnego.client(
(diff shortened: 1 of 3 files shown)


jborean93:
It'll be better to just add a new kwarg where you can specify the password. Relying on string splitting will be error prone, especially if you start having usernames/realms with `:` in them.

razzzp:
oh so just add a "password" param to the HTTPKerberosAuth.__init__?

requests_kerberos/kerberos_.py
```diff
@@ -209,14 +210,26 @@ def generate_request_header(self, response, host, is_preemptive=False):
...
-            self._context[host] = ctx = spnego.client(
-                username=self.principal,
-                hostname=kerb_host,
-                service=self.service,
-                channel_bindings=self._cbts.get(host, None),
-                context_req=gssflags,
-                protocol="kerberos",
-            )
+            # to prevent failing compare headers test, don't pass password if None
+            if self.password:
```
jborean93, line 214: No need for the branching calls, just do `password=self.password` and it being `None` will work normally. I'm not sure what you mean by

> to prevent failing compare headers test, don't pass password if None

razzzp: it was because it was failing 14 tests, in this parts that look like:
```
    assert mock_client.call_args[1] == {
        "username": None,
        "hostname": "www.example.org",
        "service": "HTTP",
        "channel_bindings": None,
        "context_req": spnego.ContextReq.sequence_detect | spnego.ContextReq.mutual_auth,
        "protocol": "kerberos",
    }
```
https://github.com/requests/requests-kerberos/blob/3f672cf7d175d7c073e3a52eeb979e33930330ee/tests/test_requests_kerberos.py#L73C1-L80C6

since when I pass password even if it is none, it seems to add an explicit `"password": None` in the args, which fails the assert.

and I was too lazy to update all the tests hehe. 
but let me know if i should just add the additional args in the tests.

jborean93: I would fix the tests, no need to make the code more complicated than it needs to be.

razzzp: alright, i've updated the tests

jborean93 (approved):

Issue #178, opened 2022-07-08
force_preemptive=True is only preemptive the first time


gconklin:
When setting force_preemptive=True, it looks like only the first request gets an immediate 200 and all other following requests get a 401, and then a 200.

I'm essentially doing this:

```
import requests
from requests_kerberos import HTTPKerberosAuth
from requests_kerberos.exceptions import KerberosExchangeError

class Thing:
    def __init__(self):
        self.session = requests.Session()
        self.session.auth = HTTPKerberosAuth(force_preemptive=True)

    def get(self, url):
        return self.session.get(url)

t = Thing()
t.get("https://hostname/")
t.get("https://hostname/")
```

```
First request:
DEBUG:requests_kerberos.kerberos_:HTTPKerberosAuth: Preemptive Authorization header: Negotiate ...
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): hostname:443
DEBUG:urllib3.connectionpool:https://hostname:443 "GET / HTTP/1.1" 200 None
DEBUG:requests_kerberos.kerberos_:handle_other(): Handling: 200
DEBUG:requests_kerberos.kerberos_:handle_other(): Authenticating the server
DEBUG:requests_kerberos.kerberos_:authenticate_server(): Authenticate header:
DEBUG:requests_kerberos.kerberos_:authenticate_server(): returning <Response [200]>
DEBUG:requests_kerberos.kerberos_:handle_other(): returning <Response [200]>
DEBUG:requests_kerberos.kerberos_:handle_response(): returning <Response [200]>

Next request:
DEBUG:urllib3.connectionpool:https://hostname:443 "GET / HTTP/1.1" 401 381
DEBUG:requests_kerberos.kerberos_:handle_401(): Handling: 401
DEBUG:requests_kerberos.kerberos_:authenticate_user(): Authorization header: Negotiate ...
DEBUG:urllib3.connectionpool:https://hostname:443 "GET / HTTP/1.1" 200 None
DEBUG:requests_kerberos.kerberos_:authenticate_user(): returning <Response [200]>
DEBUG:requests_kerberos.kerberos_:handle_401(): returning <Response [200]>
DEBUG:requests_kerberos.kerberos_:handle_response(): returning <Response [200]>
DEBUG:requests_kerberos.kerberos_:handle_response() has seen 0 401 responses
DEBUG:requests_kerberos.kerberos_:handle_other(): Handling: 200
[...]

PR #183, opened 2023-09-01
Send Authorization header irrespective of previous authentication status (#178)

chamakuri-vineel:
- Currently 'force_preemptive' flag sends the 'Authorization' header only for the first call and not for subsequent calls. This is because 'Authorization' header is preemptively sent only if the request is not authenticated. The status of authentication is stored as 'auth_done' instance variable. Since all the HTTP calls use the same instance of HTTPKerberosAuth class the 'auth_done' instance variable is set to True after first successful authentication and thus 'Authorization' header is not set preemptively after that.
- The fix is to send the Authorization header irrespective of the authentication status of previous call.


chamakuri-vineel:
@mkomitee @josecastroleon @dpursehouse  please review the PR. This is to fix #178  issue.

PR #192, opened 2024-10-11
(test) build wheels with cibuildwheel

jamesfricker:
Hoping this will ensure the user doesn't actually need the underlying packages and can pip install on linux without krb5


jborean93:
What are you trying to achieve with these changes? The PyPI package already contains a universal wheel for this library https://pypi.org/project/requests-kerberos/#files. The underlying Kerberos/GSSAPI dependency doesn’t have a wheel for Linux on PyPI for various reasons which we cannot change unfortunately.

jamesfricker:
apologies - this doesn't work how I thought, you are correct. Was hoping this would work but it seems that it's not possible. 😔 Thanks for the comment!

